PR #19, opened 2015-01-06
Get and pass up feature to update higher caches

raadad:
If higher level caches don't have a key, they will be given it if it exists in a lower level.

### DIFF
--- a/lib/multi_caching.js
+++ b/lib/multi_caching.js
@@ -43,10 +43,9 @@ var multi_caching = function (caches) {
                 return cb(err);
             }
 
-            result = result || {};
             cb(err, result);
 
-            if (result && index) {
+            if (result !== undefined && index) {
                 var cachesToUpdate = caches.slice(0, index);
                 async.forEach(cachesToUpdate, function(cache, async_cb) {
                     cache.set(key, result, result.ttl, async_cb);
--- a/lib/multi_caching.js
+++ b/lib/multi_caching.js
@@ -31,6 +31,30 @@ var multi_caching = function (caches) {
         }, cb);
     }
 
+     /**
+     * Looks for an item in cache tiers.
+     *
+     * When a key is found in a lower cache, all higher levels are updated
+     */
+
+    self.get_and_pass_up = function(key, cb) {
+        get_from_highest_priority_cache(key, function(err, result, index) {
+            if (err) {
+                return cb(err);
+            }
+
+            result = result || {};
+            cb(err, result);
+
+            if (result && index) {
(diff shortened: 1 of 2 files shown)


BryanDonovan:
Thanks for this pull request.  Can you please add tests like:

```
            context("when value is not found in any cache", function() {
                var response;

                beforeEach(function(done) {
                    key = support.random.string(10);
                    sinon.spy(memory_cache, 'set');
                    sinon.spy(memory_cache2, 'set');
                    sinon.spy(memory_cache3, 'set');

                    multi_cache.get_and_pass_up(key, function (err, result) {
                        check_err(err);
                        response = result;
                        done();
                    });
                });

                afterEach(function() {
                    memory_cache.set.restore();
                    memory_cache2.set.restore();
                    memory_cache3.set.restore();
                });

                it("calls back with null", function() {
                    assert.strictEqual(response, null);
                });

                it("does not set anything in caches", function(done) {
                    process.nextTick(function () {
                        assert.ok(memory_cache.set.notCalled);
                        assert.ok(memory_cache2.set.notCalled);
                        assert.ok(memory_cache3.set.notCalled);
                        done();
                    });
                });
            });

```

The second test passes now, which is great.  I just want to prevent a future bug there.  The first test fails though.  I don't think it should be setting the result to `{}` when nothing is returned.

BryanDonovan:
I'll get this in a release ASAP.

BryanDonovan:
This is in release 0.16.0.  Thanks again for the pull request!

BryanDonovan:
@raadad - I'm doing some refactoring of the library and noticed that you're passing `result.ttl` to `cache.set()` inside `get_and_pass_up()`.  Are you depending on `result.ttl` existing?  I'd like to remove that since there's no guarantee that the `result` object will have a `ttl` property.  I was thinking we could pass an `options` object into `get_and_pass_up()` if there's a need to specify a `ttl` in that function.

```javascript
async.each(cachesToUpdate, function(cache, next) {
    cache.set(key, result, result.ttl, next);
});
```

Let me know what you think.

raadad:
@BryanDonovan the interface for cache.set is (key, value, ttl, cb)  - although this is ignored in memory.js it is not in redis_store.js.  I think changing ttl to options is a good idea , however all implementations of cache  will need to be refactored aswell, which will create a backwards incompatibility for custom implementations (not a big change though)

BryanDonovan:
I mean the result object does not have a ttl property.

raadad:
Oh, sorry for misunderstanding, yes, you are right.

PR #20, opened 2015-02-02
Add Additional Options Parameter

seanzx85:
Add the ability to pass in options.  These options can then be used by the stores.
Our use case is using Riak as a distributed fault tolerant cache.
We want to pass in if quorum is required for writes, read or deletes.

A get request would look like:
```
cache.get('ActiveLock', function(err, val){}, {quorum: true});
```

Also there is a test case fixed in ef044b8  The timeout parameter was on the wrong closing curly brace.

### DIFF
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -1,4 +1,5 @@
 /*jshint maxcomplexity:15*/
+/*jshint -W072 */
 var domain = require('domain');
 
 var caching = function(args) {
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -36,9 +36,10 @@ var caching = function(args) {
      *       console.log(user);
      *   });
      */
-    self.wrap = function(key, work, ttl, cb) {
+    self.wrap = function(key, work, ttl, cb, options) {
         if (typeof ttl === 'function') {
             cb = ttl;
+            options = cb;
             ttl = undefined;
         }
 
@@ -79,10 +80,10 @@ var caching = function(args) {
                         } else {
                             fillCallbacks(null, data);
                         }
-                    });
+                    }, options);
                 });
             }
-        });
+        }, options);
     };
 
     self.get = self.store.get.bind(self.store);
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -37,11 +37,10 @@ var caching = function(args) {
      *       console.log(user);
      *   });
      */
-    self.wrap = function(key, work, ttl, cb, options) {
-        if (typeof ttl === 'function') {
-            cb = ttl;
-            options = cb;
-            ttl = undefined;
+    self.wrap = function(key, work, options, cb) {
+        if (typeof options === 'function') {
+            cb = options;
+            options = undefined;
         }
 
         if (self.queues[key]) {
@@ -59,7 +58,7 @@ var caching = function(args) {
             delete self.queues[key];
         }
 
-        self.store.get(key, function(err, result) {
+        self.store.get(key, options, function(err, result) {
             if (err && (!self.ignoreCacheErrors)) {
                 fillCallbacks(err);
             } else if (result) {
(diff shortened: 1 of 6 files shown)


BryanDonovan:
Thanks for the pull request.  I'll check it out in more detail as soon as possible.

seanzx85:
Just a heads up.  I had to relax the JSHint options for the number of parameters.  I didn't want to break backwards compatibility and turn the key into an object.

BryanDonovan:
Which node riak client are you using?

seanzx85:
We have been using https://github.com/nlf/riakpbc with pretty good performance.  It is a bit heavy on the CPU side for large imports of records.

BryanDonovan:
I'm a bit hesitant to add an options param after the callback.  It kind-of goes against the node convention of having the callback be the last param.  I don't see where these options are passed in like this in RiakPBC.. can you please point me to that?  I just see code like:

```
RiakPBC.prototype.get = function (params, callback) {

    return this._makeRequest({
        type: 'RpbGetReq',
        params: params,
        callback: callback
    });
};
```

seanzx85:
We wrote our own wrapper to RiakPBC do the the complexities of the environments (such as buckets and key spacing) and types of data stored.

BryanDonovan:
Hmm.  Do you need the ability to pass different options dynamically?  Just trying to understand why you can't set the `quorum` value once in a client and wrap the other functions to use it.  Something like:

```javascript
function RiakCache(args) {
    this.quorum = args.quorum;
    /// etc.
}

RiakCache.prototype.get = function(args, cb) {
    MyRiakWrapper.get(args, cb, {quorum: this.quorum});
};

```

seanzx85:
So for example:

I want to write 20K JSON objects that are not critical.  I would write them to the cluster with a r of 1.  (Basically fire and forget).  

Lets say I need to performance a 5 minute task on one server from the cluster.  I write a "lock" into the bucket with a w of "quorum".  Then I read the "lock" value with a r of "quorum" as well. If I need to guarantee that "lock" is deleted, it would have a w of "n".

We do have global level configs but need the ability to dynamically set them.

BryanDonovan:
OK.  I think I follow.  I'm not sure I get why you can't create a new cache client instance for each quorum level though.  

I take it there's no way to move the options in your code to be before the callback instead of after?

seanzx85:
My logic was that by putting it at the end it doesn't break compatibility. 

I can move it in front of the callback and then do the typeof === function or typeof === object to still make options an optional part of the code.

It is your call because it is your library :-)

BryanDonovan:
Ah, I see.   Yeah, I guess it's better as you have it.  I wish I would have made the ttl param an object instead of a number, then we could just use that.

seanzx85:
that is another option.  We could switch that out to an object.

BryanDonovan:
I'll think about it a bit.  Don't want to break other people's code using that ttl either.  But maybe it's worth it.

BryanDonovan:
I think we should make the current `ttl` arg an object for these options.  Maybe we could preserve backward compatibility by checking if `options` is an object, and if not (and it's not the callback function), assume it's the ttl.  

Do you want to take a crack at this, or should I?

seanzx85:
I can start working on it today and see if I can get it in for tomorrow.  How do we handle GETs and DELs?  They don't have a ttl object.

BryanDonovan:
For GET/DEL, I guess we can just add the options param before the callback.

seanzx85:
Pushed up the changes and test case updates.  Only issue I see is, if you were using the memory store and a custom store that need options.  Memory store would be confused my the object instead of value.  Do you want to leave it that way or update the memory store as well?

BryanDonovan:
This looks good.  I think we should update the memory store to take options as well, and just ignore them. 

Only thing left I think is that caching.js still has options as the last param.

seanzx85:
Updated caching.js and added some more test cases to check memory ignoring the object vs the ttl

BryanDonovan:
Looks great.  I'll merge and cut a release ASAP.

BryanDonovan:
This is in 0.17.0 release now.  Thanks again!

PR #22, opened 2015-03-29
Pass dispose, length & stale options to lru-cache

gmaclennan:
The length option is particularly useful with lru-cache if you want to store large objects of varying size and limit memory usage.
https://github.com/isaacs/node-lru-cache#options


BryanDonovan:
Thanks!  This is available now in 0.19.0

PR #27, opened 2015-05-21
Fix redis store TTL

homaily:
Hello Bryan,

On redis store example when you set custom TTL with ``set`` function.

It doesn't work because ``set`` function check for ``options.ttl`` key and it doesn't check if the options parameter is numeric.


BryanDonovan:
Hi @homaily,

Thanks for the pull request.  However, I'm in the process of removing support for the numeric ttl so that the caching code is simpler and can be more easily applied to different cache backends.

PR #29, opened 2015-06-01
Stores can override isCacheableValue

PuKoren:
Potential issue was instroduced on https://github.com/BryanDonovan/node-cache-manager/commit/f3809c8b762b38e8ce0a22155b528e419b5602fc using Redis stores and wrap.
This was done following the bug report https://github.com/BryanDonovan/node-cache-manager/issues/25 I'm sorry I was not able to confirm the fix so I made this PR

Using Redis, null is returned from cache when data does not exists
In other stores, null can still be stored and undefined used when data does not exist

The issue occurs only when calling wrap in simple cache, not multi cache so I guess multi cache do not use isCacheableValue

This PR includes:
  - Each store can implement its own custom method
  - Tests to keep coverage to 100%
  - Updated redis cache exemple with the override since redis can return null when data does not exists

### DIFF
--- a/examples/redis_example/redis_store.js
+++ b/examples/redis_example/redis_store.js
@@ -110,6 +110,10 @@ function redisStore(args) {
         });
     };
 
+    self.isCacheableValue = function(value) {
+        return value !== null && value !== undefined;
+    };
+
     return self;
 }
 
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -34,6 +34,8 @@ var caching = function(args) {
 
     if (typeof args.isCacheableValue === 'function') {
         self._isCacheableValue = args.isCacheableValue;
+    } else if (typeof self.store.isCacheableValue === 'function') {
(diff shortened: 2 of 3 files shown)


PuKoren:
I was not able to implement it in a clean way for multi caching.
Another solution could be to decide to not cache null or undefined instead of just undefined for every cache.

BryanDonovan:
Thanks, I'll check this out in more detail when I have some time.  In the meantime, you can just pass in your own `isCacheableValue` function when initializing `caching` or `multiCaching`

```javascript
var isCacheableValue = function(value) {
    return value !== null && value !== undefined;
};

var memoryCache = cacheManager.caching({store: 'memory', isCacheableValue: isCacheableValue};

var multiCache = cacheManager.multiCaching([memoryCache, someOtherCache], {
    isCacheableValue: isCacheableValue
});
```

BryanDonovan:
Sorry this took so long to merge.  I'll get it in a new release ASAP.

Issue #24, opened 2015-05-11
Support promises


Jokero:
Native promises are already supported in nodejs 0.12. I think all methods should return promise. And `wrap` method can take function returning promise:
```javascript
/**
 * @returns {Promise}
 */
function getCachedUser(id) {
    return memoryCache.wrap(id, function(/* optional cacheCallback */) {
        return getUser(id);
    }, ttl);
}
```

BryanDonovan:
Thanks @Jokero - do you want to try implementing this?

standy:
+1 on that feature

BryanDonovan:
I don't use promises myself, so if someone else wants to try implementing support for them, that would be great.

cgmartin:
In case it helps anyone, here's an example workaround using bluebird's `promisifyAll()` with cache-manager's `wrap()` function: https://gist.github.com/cgmartin/2df6f68d08388bc8c398

(The first `work/cacheMiss` callback doesn't precisely match @Jokero's original example, but it gets close. I'd be curious if that could be improved upon.)

I'm not sure if it was explicitly called out, but if promise support was added consider making it optional for those who don't use promises... using something like [bluebird's `.asCallback()`](http://bluebirdjs.com/docs/api/ascallback.html)
> This can be used to create APIs that both accept node-style callbacks and return promises

jkernech:
+1

BryanDonovan:
It does seem like using `Promise.promisifyAll` or equivalent would work, right? I could export a "promisified" version of the module perhaps.  I'm not sure what library to use though (I don't see built-in promises in node 0.12.x).

marttiku:
My solution for the problem using bluebird.

Setup:

```javascript
var cacheManager = require('cache-manager');
var cache = cacheManager.caching({
    store: 'memory',
    max: 100,
    ttl: 30 /*seconds*/
});

bluebird.promisifyAll(cache);

cache.promise = function (cacheKey, work, ttl) {
	ttl = ttl || 10;
    return cache.wrapAsync(cacheKey, function(cacheResult) {
            return work().asCallback(cacheResult);                   
        }, {
            ttl: ttl
        }
    );	
}
```

and usage:

```javascript
            var cachekey = 'session_' + req.token;
            var ttl = 10;

            function work() {
                return models.Session.findOne({
                    where: {
                        token: req.token
                    }
                })                
            }

     var promise =  cache.promise(cachekey, work, ttl);
```

[...]

PR #39, opened 2016-01-18
Promise support

PuKoren:
To handle #24 

There is the promise support WIP branch to make the cache manager handle promises.

I have some issues on the existing cache providers, and for tests. Right now only wrap (simple cache and multi cache) and get/set (multi cache) are implemented.

The issue with single store cache is that cache-manager is directly binded to the underlying methods, and to add promise support in get I need to change the method usage and add a breaking change, that I would like to avoid.

Some more tests to be added (100% on wrap, drop to 97% on get/set ATM)

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -59,7 +59,7 @@
     "expr"          : false,  // Tolerate `ExpressionStatement` as Programs.
     "forin"         : false,  // Tolerate `for in` loops without `hasOwnProperty`.
     "immed"         : true,   // Require immediate invocations to be wrapped in parens e.g. `( function(){}() );`
-    "latedef"       : true,   // Prohibit variable use before definition.
+    "latedef"       : "nofunc",   // Prohibit variable use before definition.
     "loopfunc"      : true,   // Allow functions to be defined within loops.
     "maxparams"     : 4,
     "maxdepth"      : 5,
--- a/.jshintrc
+++ b/.jshintrc
@@ -20,6 +20,7 @@
 
     "predef"        : [  // Extra globals.
         "__dirname",
+        "Promise",
         "Buffer",
         "event",
         "exports",
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -69,6 +69,17 @@ var caching = function(args) {
             options = {};
         }
 
+        if (!cb) {
+            cb = Promise.defer();
+            var work2 = work;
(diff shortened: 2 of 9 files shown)


PuKoren:
Travis should be configured to use node >= 0.11.13 with --harmony tags to make the tests pass, or a recent stable version including native Promises. A shim can be used to keep retro-compatibility with older versions of node, such as Bluebird, but this makes the dependency list grow

jkernech:
:+1: !!!

BryanDonovan:
Thanks for this.  I need to think a bit about how to go forward.  Maybe releasing a 2.0.0 version with promise support and tell anyone using older versions of node need to use 1.2.x.  I'm hesitant to do that though.

Also, is it possible for a user to just do `promisifyAll()` or equivalent to get this to work?  I.e, is it difficult for a user to get 1.2.2 to work with promises without changing the node-cache-manager code?

PuKoren:
I think a user can use promises using promisifyAll but I'm not sure about that because I don't used much libs such as bluebird.

What libs are usually doing, is returning a promise only if the callback parameter is not provided. Is it something we can imagine, so lib would stay in 1.x?

PuKoren:
Just did that, but I'm using deprecated defer() methods with promises, this should be refactored but need more work.
ATM it is compatible using callback on every node versions, and using promises on node > 0.12
However not compatible with polyfills and bluebird because of deprecated method usage

jkernech:
@BryanDonovan What do you think of this PR now ?

I think it's an awesome improvement and to be honest, I can't wait to use it this way !!

BryanDonovan:
This is looking good.  I'm not sure I understand the implications of using the deprecated `defer()` method though.

PuKoren:
Here we go: I removed the deprecated usage of defer() and used simple wrappers around functions when no callback are provided.

Now it is compatible with older versions of node, as long as users use the lib the same way as they did before: no breaking changes.

If they have node > 0.12, they have Promise support and can use the lib the same as before or with Promises.
With node < 0.12 you can use the lib with promises if you include a Polyfill, like:
```
global.Promise = require('es6-promise').Promise;
```

I think it is safe to tag it 1.3

jkernech:
:+1:

BryanDonovan:
OK, this is now in 1.3.0.  Thanks guys!

PR #42, opened 2016-01-26
Breaking out of asyncSeries loop when value is found. Controlled via new options function isReturnableValue.

3choBoomer:


PuKoren:
If you just check for ``result`` var I think you won't break out of async loop in cases where cached value is a falsey (false, "", 0, etc.)

Maybe I'm mistaken, is it expected behaviour?

3choBoomer:
@PuKoren my thought there was that the expense of getting a falsey value from a cache is extremely cheap compared to non-falsey values like large strings. (which is the specific use case I'm attempting to solve for here).

I'm working on doing this another way that came out of my discussion with @BryanDonovan here:
https://github.com/BryanDonovan/node-cache-manager/issues/41

BryanDonovan:
Sorry @3choBoomer -- I didn't see your comment here.  Can you merge the latest from upstream/master and add the same `isReturnableValue` logic to `lib/caching.js` as well please?  Then I think it's good to go.

botika:
Closed by refactor #203

Issue #38, opened 2015-12-07
This domain module is pending deprecation


jkernech:
The errors are handled via `domain` and the module is deprecated `Stability: 0 - Deprecated` : https://nodejs.org/api/domain.html

It should be replaced with a listener :
```js
process.on('uncaughtException', function (err) {
  console.log(err.stack);
})
```

BryanDonovan:
Cool.  I'll remove when I get some time (unless someone else wants to submit a pull request with the changes).

jkernech:
:+1:

zlokomatic:
I would also like to see this fixed

BryanDonovan:
@jkernech - sorry, I didn't see your message until now.  If someone else wants to do a pull request, that would be great.  Otherwise it might be a couple weeks before I can get to this.

BryanDonovan:
@jkernech, @zlokomatic - can you please check out https://github.com/BryanDonovan/node-cache-manager/pull/44 and let me know what you think?

This will probably require bumping the version to 2.0.0 since some people might be relying on `domain` catching the errors and passing them back in the callbacks.

zlokomatic:
Looks good :thumbsup:

PR #44, opened 2016-02-16
removing domain error catching from caching/multi-caching

BryanDonovan:
Removes `domain` error handling (see #38).


BryanDonovan:
Domains have been removed in version 2.0.0.

PR #48, opened 2016-04-18
Triggering callback in wrap after it was queued

theorm:
Just noticed that callback in `wrap` method never gets triggered second time if a non cacheable value is returned. This pull request fixes it.


BryanDonovan:
Thanks.  I'll merge shortly.

BryanDonovan:
Available in version 2.0.1.

theorm:
Thanks Bryan!

PR #56, opened 2016-06-01
add node-cache-manager-fs-binary storage

sheershoff:
Better performance for binary data caches than in node-cache-manager-fs


BryanDonovan:
Thanks!

PR #60, opened 2016-07-12
Add node-cache-manager-memcached-store ref

theogravity:
Add node-cache-manager-memcached-store link to readme

See https://github.com/theogravity/node-cache-manager-memcached-store


BryanDonovan:
Thanks, merging now.

PR #65, opened 2016-10-24
Feature/dynamic cache ttl

philippeauriach:
Allow users to compute TTL based on the wrapped function result data (instead a fixed value).
Needed this for a simple case of getting values back from an API, with values having an expiration date.

Usage : 

```
myCache.wrap(KEY, function(){
    return dataPromise;
  }, {
    computeTTL: function(data){
        if(isSoon(data.endOfLife)){
            return 600;
        }
        return 6000;
    }
  })
```

### DIFF
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -113,6 +113,10 @@ var caching = function(args) {
                         return;
                     }
 
+                    if(typeof options.computeTTL === 'function'){
+                      options.ttl = options.computeTTL(data);
+                    }
+
                     self.store.set(key, data, options, function(err) {
                         if (err && (!self.ignoreCacheErrors)) {
                             callbackFiller.fill(key, err);
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -115,6 +115,7 @@ var caching = function(args) {
 
                     if (typeof options.computeTTL === 'function') {
                         options.ttl = options.computeTTL(data);
+                        delete options.computeTTL;
                     }
 
                     self.store.set(key, data, options, function(err) {
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -113,8 +113,8 @@ var caching = function(args) {
                         return;
                     }
 
-                    if(typeof options.computeTTL === 'function'){
-                      options.ttl = options.computeTTL(data);
+                    if (typeof options.computeTTL === 'function') {
+                        options.ttl = options.computeTTL(data);
                     }
 
                     self.store.set(key, data, options, function(err) {
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -113,9 +113,9 @@ var caching = function(args) {
                         return;
                     }
 
-                    if (options && typeof options.computeTTL === 'function') {
-                        options.ttl = options.computeTTL(data);
-                        delete options.computeTTL;
+                    if (options && typeof options.computeTtl === 'function') {
+                        options.ttl = options.computeTtl(data);
+                        delete options.computeTtl;
                     }
 
                     self.store.set(key, data, options, function(err) {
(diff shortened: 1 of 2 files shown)


BryanDonovan:
Thanks, I'll check this out as soon as I can.

BryanDonovan:
This might be nit-picking but can you change the name `computeTTL` to `computeTtl` please?  I really prefer that style of camelcase.

philippeauriach:
sure

BryanDonovan:
I need to think about this a bit more.  I'm not sure that this `computeTtl` function should be passed into the `wrap` function.  It seems to me that it might be better to pass it to the underlying store. 

For example, a redis store's `set` function could include it like so:

```javascript
    self.set = function(key, value, options, cb) {
        if (typeof options === 'function') {
            cb = options;
            options = {};
        }
        options = options || {};

        if (options && typeof options.computeTtl === 'function') {
            options.ttl = options.computeTtl(value);
            delete options.computeTtl;
        }

        var ttl = (options.ttl || options.ttl === 0) ? options.ttl : ttlDefault;

        connect(function(err, conn) {
            if (err) { return cb(err); }
            var val = JSON.stringify(value);

            if (ttl) {
                conn.setex(key, ttl, val, handleResponse(conn, cb));
            } else {
                conn.set(key, val, handleResponse(conn, cb));
            }
        });
    };
```

Or maybe it just needs to be more generic.  For example, we could pass in an option named something like `modifyOptions`, something like:

```javascript
if (options && typeof options.modifyOptionsForSet === 'function') {
    options = options.modifyOptionsForSet(options, data);
}
```

Then manipulating the options wouldn't be restricted to `ttl` modifications.  But I'm not sure I like this pattern either...  Hmm.

philippeauriach:
If I read this correctly, the first solution implies that the enhancement should be made in each store ? It doesn't seems right as the ttl property is global to all stores.

The idea of having a more generic solution is cool, we could just add the ability to pass either an object with the options (as of today) or a function returning the option objects. That would implies maybe separating options used for 'get' and options used for 'set'.

then we could do the following : 

```
cache.wrap(key, function(cb) {
  methods.getWidget(name, cb);
}, function(data){
 //check for ttl inside data
 return {
  ttl: 12
 }:
}, function(err, widget) {
 ...
});
```

BryanDonovan:
I like the idea of passing either a function or an object where we currently pass an `opts` object.

philippeauriach:
Yes, this is definitely a problem. This would maybe mean a breaking change where options object is not optional anymore.
If we go to this direction, we should decide what to do for the "options" object passed to the "get" method. I imagine options here are different from options used to set ? Should we call the function a first time without data, then call it again with data ?

BryanDonovan:
Sorry for the delayed response.  I don't think I want to do a breaking change for this.  I suppose the original implementation will work.  I might try a couple things myself if I get some time.  Otherwise I'll merge the existing changes soon.

BryanDonovan:
What about just calling the function `ttl`?  E.g.,

```javascript
if (options && typeof options.ttl === 'function') {
    options.ttl = options.ttl(data);
}
```

I'm not sure it adds much, but I'm thinking that if someone wants to add another computable option, then this approach will be easier to make it generic.  Something like:

```javascript
if (options) {
    Object.keys(options).forEach(function(key) {
        if (typeof options[key] === 'function') {
            options[key] = options[key](data);
        }
    });
}
```

philippeauriach:
this looks like the best compromise. I'll do that ASAP

BryanDonovan:
@philippeauriach - sorry for the really long delay!  I somehow missed your last commit.  I'll merge this ASAP.

BryanDonovan:
This is available in version 2.4.0

PR #69, opened 2016-12-16
Use svg badges in readme

lukechilds:
Sorry couldn't resist, the blurriness offended me.


BryanDonovan:
Thanks, that annoyed me too.

PR #72, opened 2016-12-22
null as first argument to keep style consistent

defcc:
Add null as first argument for callback if no error.


BryanDonovan:
Thanks!

PR #94, opened 2017-05-02
Add node-cache-manager-memory-store link to readme

theogravity:
I note that `node-cache-manager` already has a memory store by default in my repo.


BryanDonovan:
thanks!

PR #96, opened 2017-05-04
Add node-cache-manager-ioredis link to readme

dabroek:
Added link to the `cache-manager-ioredis` store to README.md. This store is directly built upon `ioredis`.


BryanDonovan:
Thanks!

PR #106, opened 2017-09-20
fix multicaching when result is not cacheable

### LINKED ISSUE #105: Multicaching error when isCacheableValue() == false
The following code exits early:
```js
const cacheMan = require('cache-manager');

const cache = cacheMan.multiCaching([
  cacheMan.caching({ store: 'memory' }),
]);

function run(id) {
  return cache.wrap(id, () => Promise.resolve(undefined)).then(result => {
    console.log(`Result: %s`, result);
  });
}

console.log('First run');
run('abc').then(() => {
  console.log('Second run');
  return run('abc');
}).then(() => {
  console.log('Third run');
  return run('xyz');
});
```
Outputs:
```
First run
Result: undefined
Second run
```

While this block is successful (only change is `undefined` to `'success'`):
```js
const cacheMan = require('cache-manager');

const cache = cacheMan.multiCaching([
  cacheMan.caching({ store: 'memory' }),
]);

function run(id) {
  return cache.wrap(id, () => Promise.resolve('success')).then(result => {
    console.log(`Result: %s`, result);
  });
}

console.log('First run');
run('abc').then(() => {
  console.log('Second run');
  return run('abc');
}).then(() => {
  console.log('Third run');
  return run('xyz');
});
```
Outputs:
```
First run
Result: success
Second run
Result: success
Third run
Result: success
```

I believe the problem lies in https://github.com/BryanDonovan/node-cache-manager/blob/053296d8481224a8a73413e28d67dd415e275816/lib/multi_caching.js#L242 as the CallbackFiller is never cleared, resulting in the second run exiting here https://github.com/BryanDonovan/node-cache-manager/blob/053296d8481224a8a73413e28d67dd415e275816/lib/multi_caching.js#L223

I may attempt a fix, but wanted to log this issue in case I get sidetracked.

gswalden:
Closes #105 

If you isolate the new unit test and run it against current `master`, you'll see the test fail via timeout.

### DIFF
--- a/lib/callback_filler.js
+++ b/lib/callback_filler.js
@@ -4,7 +4,7 @@ function CallbackFiller() {
 
 CallbackFiller.prototype.fill = function(key, err, data) {
     var waiting = this.queues[key];
-    delete this.queues[key];
+    this.clear(key);
 
     if (waiting && waiting.length) {
         waiting.forEach(function(task) {
--- a/lib/callback_filler.js
+++ b/lib/callback_filler.js
@@ -25,4 +25,8 @@ CallbackFiller.prototype.add = function(key, funcObj) {
     }
 };
 
+CallbackFiller.prototype.clear = function(key) {
+    delete this.queues[key];
+};
+
 module.exports = CallbackFiller;
--- a/lib/callback_filler.js
+++ b/lib/callback_filler.js
@@ -4,7 +4,7 @@ function CallbackFiller() {
 
 CallbackFiller.prototype.fill = function(key, err, data) {
     var waiting = this.queues[key];
-    this.clear(key);
+    delete this.queues[key];
 
     if (waiting && waiting.length) {
         waiting.forEach(function(task) {
@@ -23,10 +23,6 @@ CallbackFiller.prototype.add = function(key, funcObj) {
     } else {
         this.queues[key] = [funcObj];
     }
-};
-
-CallbackFiller.prototype.clear = function(key) {
-    delete this.queues[key];
 };
 
 module.exports = CallbackFiller;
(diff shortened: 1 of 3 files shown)


gswalden:
Looks like the following test still breaks, closing till fixed.
```js
it("allows repeated calls for uncacheable value at same time", function (done) {
    function getUndefined(name, cb) {
        multiCache.wrap(key, function(cacheCb) {
            process.nextTick(function() {
                cacheCb(null, undefined);
            });
        }, cb);
    }

    var i = 0;
    function cb() {
        if (++i === 2) done();
    }

    var name = 'oof';

    getUndefined(name, function(err, val) {
        checkErr(err);
        assert.ok(val === undefined);
        cb();
    });
    getUndefined(name, function(err, val) {
        checkErr(err);
        assert.ok(val === undefined);
        cb();
    });
});
```

gswalden:
Okay, simpler code changes and issue fixed.

theorm:
Just noticed that exact same fix I made two years ago never made it into the release although it is marked as merged: https://github.com/BryanDonovan/node-cache-manager/pull/48

BryanDonovan:
@gswalden - sorry, I didn't notice your bug or fix.  I was out of town that weekend and I guess I skimmed over my emails too quickly.  I'll take a look ASAP.

@theorm - I think that fix was in the `lib/caching.js` file, not in the `muliticaching.js` file. (https://github.com/BryanDonovan/node-cache-manager/pull/48/commits/e9c4f0af2a7e1f14eeb91691a8dcbb757f03f662).

BryanDonovan:
This is fixed in 2.6.0
CC @yoavbaruch, @rubenstolk

PR #111, opened 2017-11-30
Add human readable ttl support

Cap32:
Using [ms](https://github.com/zeit/ms) to parse ttl if ttl is a string.

Examples: `ttl: '1 day'`, `ttl: '3.2 hours'`.

### DIFF
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -1,6 +1,7 @@
 /** @module cacheManager/caching */
 /*jshint maxcomplexity:15*/
 var CallbackFiller = require('./callback_filler');
+var ms = require('ms');
 
 /**
  * Generic caching interface that wraps any caching library with a compatible interface.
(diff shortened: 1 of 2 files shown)


BryanDonovan:
Thanks for the pull request.  I'm not sure I want to add a dependency just for human-readable TTLs though.  I think most people are comfortable using milliseconds directly.

thgreasi:
@BryanDonovan isn't it actually `seconds` instead of `milliseconds`?? Or did the scale of the ttl changed since then?
See: https://github.com/BryanDonovan/node-cache-manager/blob/b3aa1d59885d5a011fd02f8c197c78f4f6235dde/test/caching.unit.js#L731
See: https://github.com/BryanDonovan/node-cache-manager/blob/b3aa1d59885d5a011fd02f8c197c78f4f6235dde/examples/redis_example/example.js#L9

BryanDonovan:
Yes @thgreasi, it's in seconds, not milliseconds.  Thanks!

PR #114, opened 2018-02-03
feature: allow setting, getting and deleting  multiple keys

sebelga:
Hello!

This PR is to allow setting, getting and deleting multiple keys at once. It works both on the caching and the multicaching. As such, the "wrap" method now accepts an unlimited number of keys.

I am aware that this is a big feature with loads of changes. I had to do some refactoring to stay under the cyclomatic complexity. Let me know all your comments.

This feature does not add any breaking change to node-cache-manager.

I opened a PR in "node-cache-manager-redis-store" to give support to this feature in redis store (https://github.com/dabroek/node-cache-manager-redis-store/pull/6)

### DIFF
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -1,6 +1,8 @@
 /** @module cacheManager/caching */
-/*jshint maxcomplexity:15*/
+/*jshint maxcomplexity:16*/
 var CallbackFiller = require('./callback_filler');
+var utils = require('./utils');
+var parseWrapArguments = utils.parseWrapArguments;
 
 /**
  * Generic caching interface that wraps any caching library with a compatible interface.
@@ -47,12 +49,12 @@ var caching = function(args) {
         return new Promise(function(resolve, reject) {
             self.wrap(key, function(cb) {
                 Promise.resolve()
-                .then(promise)
-                .then(function(result) {
-                    cb(null, result);
-                    return null;
-                })
-                .catch(cb);
+                    .then(promise)
+                    .then(function(result) {
+                        cb(null, result);
+                        return null;
+                    })
+                    .catch(cb);
             }, options, function(err, result) {
                 if (err) {
                     return reject(err);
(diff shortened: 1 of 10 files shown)


lib/utils.js
```diff
@@ -0,0 +1,16 @@
+
+var isObject = function isObject(value) {
+    return value instanceof Object && value.constructor === Object;
+};
+
+var isArray = function isArray(value) {
```
BryanDonovan: I thought all versions of node.js have `Array.isArray`.  Is there some other context where `node-cache-manager` could be used where `Array.isArray` is not defined?

test/caching.unit.js
```diff
@@ -770,6 +900,58 @@ describe("caching", function() {
                     });
                 });
             });
+
+            context("when passing serveral keys", function() {
```
BryanDonovan: Spelling: serveral

test/utils.unit.js
```diff
@@ -0,0 +1,52 @@
...
+var isObject = utils.isObject;
+
+describe('utils', function() {
+    describe('isArray()', function() {
+        it('---> ok', function() {
+            var result1 = isArray([]);
+            assert.ok(result1);
+        });
+
+        it('---> ok (withouth isArray)', function() {
```
BryanDonovan: Spelling (withouth)

test/utils.unit.js
```diff
@@ -0,0 +1,52 @@
...
+            var original = Array.isArray;
+            Array.isArray = undefined;
+            var result1 = isArray([]);
+            assert.ok(result1);
+            Array.isArray = original;
+        });
+
+        it('---> ko', function() {
```
BryanDonovan: Can you please give these tests titles that describe a bit more what they're doing?  At first I thought `ko` was a typo.

yarn.lock
```diff
@@ -0,0 +1,2052 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
BryanDonovan: Is this file intentionally added to the repo?

sebelga:
Hi!
I will respond to each comment, above but about your concern I have to admit that I have no "best" answer either. I went this way to try to avoid duplication. As there is already quiet some duplications between "caching" and "multi_caching".
On the other side, by not having "multWrap" it simplifies the api for the developers.

I let you decide that you prefer to add a "multiWrap" method I can refactor.
I would try to avoid to duplicate the "setInMultipleCaches" and "getInMultipleCaches" though.

lib/caching.js
```diff
@@ -66,68 +68,171 @@ var caching = function(args) {
...
+     * cache.wrap(key, function(cb) {
+     *     User.get(userId, cb);
+     * }, function(err, user) {
+     *     console.log(user);
+     * });
+     *
+     * // Multiple keys
+     * var key1 = 'user_123';
+     * var key2 = 'user_456';
+     * cache.wrap(key, key2, function(cb) {
```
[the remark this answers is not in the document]
sebelga: No prob I will change that.

lib/utils.js
sebelga: You're totally right, I will remove the helper.

test/caching.unit.js
sebelga: 👍

test/utils.unit.js
sebelga: 👍

test/utils.unit.js
sebelga: No problem I will change them

yarn.lock
sebelga: Well for those who use yarn it will speed up the install of the node_modules. Do you want me to remove it?

BryanDonovan: Yeah, I think it should be removed since I don't use Yarn and it would get out of date.

lib/caching.js
```diff
@@ -66,68 +68,171 @@ var caching = function(args) {
...
         }
 
         if (!cb) {
-            return wrapPromise(key, work, options);
+            args.push(work);
+            args.push(options);
+            return wrapPromise.apply(this, args);
+        }
+
+        var isMultiple = args.length > 1;
```
sebelga: If you want, from this line I could refactor (and the same in multi_caching):

```js
if (args.length > 1) {
    return wrapMultiple(args, work, options, cb);
}
var key = args;

// Put back the current "wrap" code
```

sebelga:
I made the changes and left the commits so you can review the changes. I will squash them when all is done.
What do you think about the refactor I mention in caching.js above?

lib/caching.js
BryanDonovan: Yeah, I think I like this refactor.

sebelga:
Hi, I just did the refactor of the "wrap" function both in caching and multi_caching, adding a multiWrap when passing multiple keys. I squashed the commit so you can merge if its ok.
Let me know if there is anything else. 👍

BryanDonovan:
Thanks for the new changes.  I'll review as soon as I can, hopefully today.

lib/caching.js
```diff
@@ -130,20 +156,120 @@ var caching = function(args) {
         });
     };
 
+    function wrapMultiple(keys, work, options, cb) {
+        /**
+         * We create a unique key for the multiple keys
+         * by concatenating them
+         */
+        var uuidKey = keys.reduce(function(acc, k) {
```
BryanDonovan: Maybe call this `combinedKey` or something... `uuid` implies that it's and actual `uuid` (per RFC spec, etc.)

sebelga: Ok, I will change it.

sebelga:
All changes are done :) Let me know of anything else.

BryanDonovan:
Sorry for the delay.. been busy with work.  I'll get to this as soon as possible.  Might be Monday though.

sebelga:
No worries :)

sebelga:
Hi!
I have been working this weekend quiet some time with cache manager and I found the need to be able to set the TTL value by context (depending on the store) in multi_caching. I just added a commit for this small feature.
Let me know what you think. 👍

BryanDonovan:
Thanks for this.  I've merged and will make a new release asap.

BryanDonovan:
These features are available in 2.7.0

sebelga:
Great! thanks :+1:

Issue #115, opened 2018-02-23
multiCache.set does not set the value


rolandstarke:
Calling  `multiCache.set(key, value, options)` without options and a value from type object will not work.

This test code fails, I expected the assertion to be equal:

```javascript
const assert = require('assert');
const cacheManager = require('cache-manager');
const memoryCache = cacheManager.caching({store: 'memory'});
const multiCache = cacheManager.multiCaching([memoryCache]);

(async () => {
    const key = 'key';
    const value = {some: 'value'};
    
    await multiCache.set(key, value);
    
    assert.equal(value, await multiCache.get(key));
})().then(console.log, console.error);

```
(try at [RunKit](https://runkit.com/teammmate/5a90a02c0d8eb10012cc8ec7))

Setting the value to `'abc'` works fine. So its a problem with setting an object.

I think the problem comes from this lines: https://github.com/BryanDonovan/node-cache-manager/blob/e957821213fa1dc79620f998c99848a09ddd9445/lib/multi_caching.js#L212-L214

BryanDonovan:
Thanks @rolandstarke.  @sebelga, can you please take a look?  I think this is a regression introduced in the recent changes that allow setting multiple keys (https://github.com/BryanDonovan/node-cache-manager/pull/114).

sebelga:
Sorry... my mistake! :) I just created a PR that should fix this issue: https://github.com/BryanDonovan/node-cache-manager/pull/116

BryanDonovan:
Fix is in 2.8.0

PR #116, opened 2018-02-24
fix(#115): multiCache.set does not set the value

sebelga:


BryanDonovan:
Thanks!

BryanDonovan:
Fix is in 2.8.0

PR #117, opened 2018-02-28
fix(multi_caching): getFromHighestProrityCache()

sebelga:
The "keysToFetch" in getFromHighestProrityCache() was not being emptied causing the fetch to go through all the caches even though the result had been found.


BryanDonovan:
Thanks

BryanDonovan:
Fix is in 2.8.0

PR #118, opened 2018-03-05
Updated deps

josser:
only .each and .eachSeries are used from async, and I don't see braking changes introduced between 1.5.2 and 2.6.0 in those methods


BryanDonovan:
Thanks!  I'll merge this soon.

BryanDonovan:
This was fixed a while back

PR #119, opened 2018-03-10
Don't brake store's this context for isCacheableValue method

alexandrusavin:
Using `cache-manager-memcached-store` store's `isCacheableValue` method will fail because when caching assigns the value to `self._isCacheableValue` it does not pass the `this` context.

This PR takes care that store's context is passed to `isCacheableValue`.


BryanDonovan:
Thanks!  I'll get this merged soon.

BryanDonovan:
Sorry for the delay on this.

BryanDonovan:
This is available in 2.9.0

PR #126, opened 2019-05-25
bind isCacheableValue to it's underlying store

arash16:
We have a problem using multi-caching with cache-manager-memcached-store.
https://github.com/arash16/nuxt-ssr-cache/issues/7


BryanDonovan:
Thanks for this.  I've published this fix in 2.9.1.

PR #129, opened 2019-07-01
Development store "none"

R3VoLuT1OneR:
Store can be used for development functionality, when we want to
disable real cache, but still want to have callbacks\promises working.

It can fix the issue when we want to disable\enable the cache:
https://github.com/BryanDonovan/node-cache-manager/issues/92

### DIFF
--- a/README.md
+++ b/README.md
@@ -411,6 +411,9 @@ var multiCache = cacheManager.multiCaching([memoryCache, someOtherCache], {
 
 ```
 
+### Development environment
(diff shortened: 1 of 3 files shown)


BryanDonovan:
Thanks for this PR, and sorry for the delayed response (I've been on vacation).  I'll take a closer look as soon as I can.

BryanDonovan:
This is available in 2.10.0.  Thanks!

PR #133, opened 2020-02-03
Test object with prototype

ryanbecker:
This unit test currently fails, but will pass when the cached object is either not cloned or cloned with `lodash.cloneDeep`.


BryanDonovan:
Thanks!  I'll look into this asap.

BryanDonovan:
The problem is that a different test ends up failing (the one that checks that a basic function is left in tact).  Hopefully there's a way to make all situations work.

BryanDonovan:
Oops, when I tested this I didn't put the call to `cloneDeep` after the check for null.  Works now.

test/stores/memory.unit.js
```diff
@@ -79,6 +79,23 @@ describe("memory store", function() {
                 }, opts, cb);
             }
 
+            function getCachedObjectyWithPrototype(name, cb) {
```
BryanDonovan: Minor: There's a stray `y` after `Object` in the function name.

BryanDonovan:
Hmm.. this fails though:

```javascript
            function getCachedNestedFunction(name, cb) {
                cache.wrap(key, function(cacheCb) {
                    cacheCb(null, {
                        fizzle: function() {
                            return 'foo';
                        }
                    });
                }, opts, cb);
            }
//[...]
           it("preserves nested functions", function(done) {
                getCachedNestedFunction('foo', function(err, result) {
                    checkErr(err);
                    assert.equal(typeof result, 'function');

                    getCachedNestedFunction('foo', function(err, result) {
                        checkErr(err);
                        assert.equal(typeof result.fizzle, 'function');
                        done();
                    });
                });
            });
```

BryanDonovan:
Thanks for the fix @ryanbecker!

PR #135, opened 2020-02-05
adding way to disable cloning of values in memory store

BryanDonovan:
This is to allow the user to choose whether the memory store clones values before storing them in cache (via the `set` method).  Related issues:

- https://github.com/BryanDonovan/node-cache-manager/issues/100
- https://github.com/BryanDonovan/node-cache-manager/pull/103
- https://github.com/BryanDonovan/node-cache-manager/issues/132
- https://github.com/BryanDonovan/node-cache-manager/issues/134

I'm trying to decide if this change warrants a major version bump.  It makes sense to do a major bump in the sense that anyone using version 2.10.1 or earlier could run into issues with 2.10.2, 2.11.0, or 2.11.1, and the new version.  So maybe a clean break is good.  On the other hand, the new changes work the same as 2.11.1, so in that sense it's a minor bump.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,8 @@ var cacheManager = require('cache-manager');
 var memoryCache = cacheManager.caching({store: 'memory', max: 100, ttl: 10/*seconds*/});
 var ttl = 5;
 // Note: callback is optional in set() and del().
+// Note: memory cache clones values before setting them unless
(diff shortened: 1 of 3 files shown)


BryanDonovan:
This is available in 3.0.0.  To use, add `shouldCloneBeforeSet` in the `caching` options.

Issue #122, opened 2019-02-04
memory store - updateAgeOnGet 


Java-Superman:
It would be nice if the lib/stores/memory.js was changed to support updateAgeOnGet from the lru-cache.
see https://www.npmjs.com/package/lru-cache

BryanDonovan:
Hi, I'm a bit a swamped at the moment, but would gladly accept a pull request.  Otherwise I might be able to get to this in the next few weeks.  Thanks!

BryanDonovan:
Fixed in #136

PR #136, opened 2020-02-24
Added the updateAgeOnGet option

BastianKubaile:
As requsted by #122, updateAgeOnGet can now be set as a args attribute when calling cacheManager.caching(). Default is false, as is in  [lru-cache](https://github.com/isaacs/node-lru-cache/blob/master/index.js).


BryanDonovan:
Thanks for this! Available now in version 3.1.0

Issue #93, opened 2017-04-27
Can a new value be calculated before the old expired?


tkarls:
I have an app that has a function that takes a second or two to calculate (querying external servers and processes).

I'm using the cache-manager today and when accessing my app I get:

Slow (initial call. Of course it will be slow)
fast (since in cache now)
fast
fast
fast
fast
slow (since value expired)
fast
fast
fast
fast
slow

etc.

Now, is it somehow possible to start calculating the new value while still returning the old one. Like if I could set up a TTL = 10 seconds. New Calc = 8 seconds. That would give the user calling the function on second 9 a fast reply since value is still in cache but it would also trigger a recalculation so the user calling at second 11 also gets a fast response.

Once the new value is calculated it will replace the old one even if the full ttl isn't expired yet.

Can this be achieved somehow today with multi level caching or some other way?

raadad:
I wrote this to facilitate this and some othe features:

https://github.com/raadad/node-cachemire/blob/master/index.js

raadad:
Basically looking for options.cacheRefreshThreshold

BryanDonovan:
Hmm.. I'm not sure how to go about implementing this, but I'm happy to accept a pull request.

raadad:
I can try PR the following features:

// IMMENENT EXPIRY UPDATE * the feature being requested here
// UPSTREAM COALESSING 
// STAGGERED TTL

I wrote a wrapper library for cache manager, but I think people could benefit from these features

tkarls:
Interesting @raadad 
I might take a look at your wrapper. Although if you are able to PR the feature here and its accepted than that would be just super!

PR #138, opened 2020-03-13
Background refresh of expiring entries

marcoreni:
Fixes: #93 

This PR implements a mechanism to update the cached value in background by adding a new "refreshThreshold" configuration parameter for cache stores.

If `refreshThreshold` is set and if the `ttl(key)` method is available in the used store, after retrieving a value from cache the TTL of that entry will be checked.
If the remaining TTL is less than `refreshThreshold`, the system will spawn a background worker to update the value, following same rules as standard fetching. In case of multicaching, the value will be then updated in all the stores. In the meantime, the system will return the old value until expiration (or update).
In case of multicaching, the store that will be checked for refresh is the one where the key will be found first (highest priority).

NOTES:
* Background workers are put in a queue so that we avoid spawning multiple workers for the same key.
* If the `refreshThreshold` is low and the backend is slow, a racing condition may be encountered where the refresh is still. waiting to finish but at the same time a "standard" worker starts.
* The caching store needs to provide the `ttl` method (eg. standard memoryStore is not supported)
* The background refresh mechanism currently does not support providing multiple keys to `wrap` function. This is because I found it hard to find a sane way to handle this (also due to https://github.com/BryanDonovan/node-cache-manager/issues/130 )
* This flow is similar to `stale-while-revalidate` mechanism but works backwards: while in stale-while-revalidate a cache entry is kept *after* expiration for a specified period, this mechanism works *before* expiration in a specified interval. This is because otherwise we would have had to rewrite all stores to support this functionality.

Inspiration:
- https://github.com/raadad/node-cachemire
- https://github.com/cdhorne/node-cache-manager/pull/1

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,10 +432,14 @@ Both the `caching` and `multicaching` modules support a mechanism to refresh exp
 This is done by adding a `refreshThreshold` attribute while creating the caching store.
 
 If `refreshThreshold` is set and if the `ttl` method is available for the used store, after retrieving a value from cache TTL will be checked.
-If the remaining TTL is less than `refreshThreshold`, the system will spawn a background worker to update the value, following same rules as standard fetching. In case of multicaching, the value will be then updated in all the stores. In the meantime, the system will return the old value until expiration (or update).
+If the remaining TTL is less than `refreshThreshold`, the system will spawn a background worker to update the value, following same rules as standard fetching. In the meantime, the system will return the old value until expiration.
+
+In case of multicaching, the store that will be used for refresh is the one where the key will be found first (highest priority). The value will then be set in all the stores.
 
 NOTES:
 
+* In case of multicaching, the store that will be checked for refresh is the one where the key will be found first (highest priority).
+* If the threshold is low and the worker function is slow, the key may expire and you may encounter a racing condition with updating values.
 * The background refresh mechanism currently does not support providing multiple keys to `wrap` function.
 * The caching store needs to provide the `ttl` method.
 
--- a/README.md
+++ b/README.md
@@ -448,10 +448,10 @@ For example, pass the refreshThreshold to `caching` like this:
 ```javascript
 var redisStore = require('cache-manager-ioredis');
 
-var memoryCache = cacheManager.caching({store: redisStore, refreshThreshold: 3, isCacheableValue: isCacheableValue});
+var redisCache = cacheManager.caching({store: redisStore, refreshThreshold: 3, isCacheableValue: isCacheableValue});
 ```
 
-When a value will be retrieved from Redis with a TTL minor than 3sec, the value will be updated in background.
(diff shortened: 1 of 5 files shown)


BryanDonovan:
Thanks for this! I'll take a closer look ASAP.  So far it looks good.

README.md
```diff
@@ -426,6 +426,33 @@ var multiCache = cacheManager.multiCaching([memoryCache, someOtherCache], {
...
+* If the threshold is low and the worker function is slow, the key may expire and you may encounter a racing condition with updating values.
+* The background refresh mechanism currently does not support providing multiple keys to `wrap` function.
+* The caching store needs to provide the `ttl` method.
+
+For example, pass the refreshThreshold to `caching` like this:
+
+```javascript
+var redisStore = require('cache-manager-ioredis');
+
+var memoryCache = cacheManager.caching({store: redisStore, refreshThreshold: 3, isCacheableValue: isCacheableValue});
```
BryanDonovan: Minor: This should probably be named `redisCache`.

README.md
```diff
@@ -426,6 +426,33 @@ var multiCache = cacheManager.multiCaching([memoryCache, someOtherCache], {
...
+```
+
+When a value will be retrieved from Redis with a TTL minor than 3sec, the value will be updated in background.
```
BryanDonovan: I think there's a typo with "minor" here.  Maybe you meant "remaining"?

lib/caching.js
```diff
@@ -64,6 +66,58 @@ var caching = function(args) {
...
+     * @function
+     * @name checkRefreshThreshold
+     *
+     * @param {string} key - The cache key to check.
+     * @param {function} cb
+     *
+     */
+    self.checkRefreshThreshold = function(key, cb) {
+        if (self.refreshThreshold && typeof self.store.ttl === 'function') {
+            return self.store.ttl(key, function(ttlErr, ttl) {
```
BryanDonovan: I assume we don't care about the `ttlErr` here since it's not checked?  Also if `ttl` is `undefined` or `null`, it might result in strange behavior when compared to `self.refreshThreshold`.

marcoreni: I added a check for `ttlErr` and for the type of `ttl`. My assumption is that this is done "best effort". If something goes wrong the refresh is not performed.
We may implement some sort of circuit breaking mechanism to avoid overloading the store but I think that it would be out of scope for this PR.

README.md
marcoreni: Fixed. Let me know if this is more readable.

lib/caching.js
BryanDonovan: Agree, looks good now.

BryanDonovan:
I'll get this into a release in a bit.

BryanDonovan:
This is available now in 3.2.0.  Thanks again!

PR #139, opened 2020-03-16
Background refresh fixes

marcoreni:
This is a followup to #138 (a feature is not a feature if it has no bugs... )

* fix: option object passthrough and TTL calculation in multi-caching
* fix: Remove item from background queue if entry is not expired
* optimization: avoid running checks if refreshThreshold is not set (avoids some useless function calls)

### DIFF
--- a/lib/caching.js
+++ b/lib/caching.js
@@ -67,7 +67,7 @@ var caching = function(args) {
     }
 
     function handleBackgroundRefresh(key, work, options) {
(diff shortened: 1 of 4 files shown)


BryanDonovan:
Sorry for the delay.  Thanks for this.  I'll get it merged as soon as I can.

BryanDonovan:
This is available in 3.2.1

PR #148, opened 2020-11-15
add package.json browser entry and browser.js

caub:
It allows to ignore this dependency on browser, for example using webpack target: browser or rollup node-resolve browser option


caub:
not worth, dependencies can be skipped by bundlers using 'external' options

PR #154, opened 2021-03-29
Catch uncaught exception to pass it to callback

lchenay:
Some error of underlying store, like memcached-plus, throw exception that is not capture and could propagate to main loop.
CF assertKey for exemple: https://github.com/victorquinn/memcache-plus/blob/master/lib/client.js#L16-L22.

Surround by try catch to capture and restore the error to the chain promise / callback.


atondelier (approved):

fazzabi (approved):

BryanDonovan:
Thanks, I'll release this soon.

BryanDonovan:
This is available in version 3.4.2

PR #171, opened 2021-11-11
feat: add dump and load methods for memory store

jeoy:
PR for issue [How to use dump and load API since built-in memory cache is node-lru-cach](https://github.com/BryanDonovan/node-cache-manager/issues/169)

Add dump and load methods for memory store


BryanDonovan:
Thanks, looks good.  I'll get this released as soon as I can.

PR #173, opened 2021-11-11
feat: provid method to keys count

css-master:


lib/caching.js
```diff
@@ -365,6 +365,14 @@ var caching = function(args) {
         self.keys = self.store.keys.bind(self.store);
     }
 
+    /**
```
BryanDonovan, line 368: I don't think we should add this method here because it's not a common across various caches. E.g., redis uses `DBSIZE`, memcache uses `stats items`, etc.  I think this is a case where it's better to call the method directly on the store's underlying cache.  E.g., `memoryCache.store.keyCount()`, etc.

lib/stores/memory.js
```diff
@@ -186,6 +186,10 @@ var memoryStore = function(args) {
         }
     };
 
+    self.curCount = function() {
```
BryanDonovan, line 189: Can we call this `self.keyCount` instead?  Also, please add a test for this.

BryanDonovan (changes requested):

lib/caching.js
css-master: done

BryanDonovan (approved):

BryanDonovan:
This is available in version 3.6.0

PR #174, opened 2021-11-22
fix: memoryCache.get should return copy

coder-pm:


BryanDonovan:
Thanks.  I think this should use a flag similar to `shouldCloneBeforeSet` (we could call it `shouldCloneAfterGet`) and would need a couple tests.

coder-pm:
It clearly looks like a bug and shouldn't use flag. Because of that memory cache has different behaviour than other adapters. When you push object A into the cache and then modify it, it's going to change the cached instance without settting it again through the cache interface. That's uncommon for the caches and I would never expect such a behaviour.

BryanDonovan:
It's not really a bug.  If this is a bug, then the bug also exists in the LRU-cache library.  It's arguably not a good practice to be storing javascript objects in cache anyway, but if you are going to store them, you may actually *want* to be able to modify the object while it's in cache. I wouldn't do it, but someone might.

I think the default behavior of the `set` method (like the `get` method) should be not modifying the objects -- that is, the cache is a transparent wrapper around the underlying store, so users can expect the same behavior as if they used the LRU-cache library directly.

In any case, we already have `shouldCloneAfterGet`, so we should be consistent and use a flag for the `set` method as well.

BryanDonovan:
See also: https://github.com/BryanDonovan/node-cache-manager/pull/103 and https://github.com/BryanDonovan/node-cache-manager/issues/137

botika:
Closed by refactor #203

PR #187, opened 2022-04-10
add support for memory store mget to be called with array

regevbr:
Allow the same behavior as in `del()`


BryanDonovan:
Sorry for the long delay on this. I just didn't notice it for some reason.  I'll get to it soon..

botika:
Fixed in #203

PR #193, opened 2022-05-27
Upgrade lru-cache to v7

orgads:
Fixes #189.

PR #194, opened 2022-06-03
fix: ttl not working

anchan828:
Hi, I upgraded to cache-manager@4.0.0 and TTL was not working.

Perhaps it was an omission in #193 updates.

* In lru-cache@7.x, `maxAge` was renamed to `ttl`.
* The `set()` function takes options objects instead of ttl integer.

See: https://github.com/isaacs/node-lru-cache/blob/main/CHANGELOG.md#specific-api-changes

## Versions

cache-manager: 4.0.0
lru-cache: 7.10.1
node: 18.2.0
npm: 8.9.0

### DIFF
--- a/lib/stores/memory.js
+++ b/lib/stores/memory.js
@@ -47,7 +47,7 @@ var memoryStore = function(args) {
         var length = keysValues.length;
         var values = [];
         for (var i = 0; i < length; i += 2) {
-            lruCache.set(keysValues[i], keysValues[i + 1], { ttl });
+            lruCache.set(keysValues[i], keysValues[i + 1], {ttl});
             values.push(keysValues[i + 1]);
         }
         return values;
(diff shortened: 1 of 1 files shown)


anchan828:
The some tests are failing, so I will fix them.

anchan828:
Test passed😀
4ea193e16177d087dd0cc4a6955e8928

BryanDonovan:
Thanks.  I'll get this merged ASAP.

BryanDonovan:
this is available in version 4.0.1

taoliujun:
Thanks. I discover that cache get data was `undefined` a few days ago and i don't know why. I must fix this problem because app should be deploy today and i found `ttl` param was lost in lru which reason is it accept object options but not number.

Thanks your merge request!!!

PR #196, opened 2022-07-03
feat: added support for ttl as the third arg in set function of memory store

ZirionNeft:
Added support for TTL seconds as the third argument in set() function of memory store strategy.
According to types from @types/cache-manager this function should support TTL in third arg but it can't. So this PR is fix this issue.
[Link to set() type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9b425ba3f0507dd34b52cfa39772b40975fc41ef/types/cache-manager/index.d.ts#L64``)

### DIFF
--- a/test/stores/memory.unit.js
+++ b/test/stores/memory.unit.js
@@ -37,6 +37,23 @@ describe("memory store", function() {
         });
     });
 
+    describe("ttl", function() {
+        var memoryCache;
+
+        beforeEach(function() {
+            memoryCache = memoryStore.create({noPromises: true, ttl: 0.001});
+        });
+
+        it("if options arg is a number in set()", function(done) {
+            memoryCache.set('foo', 'bar', 60);
+
+            setTimeout(function() {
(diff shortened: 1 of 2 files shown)


test/stores/memory.unit.js
```diff
@@ -37,6 +37,23 @@ describe("memory store", function() {
...
+        beforeEach(function() {
+            memoryCache = memoryStore.create({noPromises: true, ttl: 0.001});
+        });
+
+        it("if options arg is a number in set()", function(done) {
+            memoryCache.set('foo', 'bar', 60);
+
+            setTimeout(function() {
+                assert.equal(memoryCache.get('foo'), 'bar');
+                done();
```
BryanDonovan, line 52: Maybe add another test here to make sure the cache record expires after the TTL?

ZirionNeft:
@BryanDonovan Additional unit test added 👍

BryanDonovan:
This is available in 4.1.0

PR #203, opened 2022-09-05
refactor: migrate to Promise and Typescript

### LINKED ISSUE #198: Migrate to typescript and Promise
It's been a long time since node has Promise support. Can I migrate the project to typescript and Promise and do the PR?

botika:
Fixes #198 

## Breaking changes 
- Remove `ignoreCacheErrors`: not valid in async/await context
- Remove `refreshThreshold`: use lower ttl 
- Remove callbackFiller 
- Use function as `FactoryStore`
- Remove multiple wrap


botika:
Finally finished, it's just a first version. Any change will be well received. I await the review.

botika:
I would like to merge it in 1st October. Please if you find any objection, comment it. Thanks

src/caching.ts
```diff
@@ -0,0 +1,99 @@
...
+  T extends object = never,
+>(store: FactoryStore<S, T>, args?: FactoryConfig<T>): Promise<Cache<S>>;
+
+/**
+ * Generic caching interface that wraps any caching library with a compatible interface.
+ */
+export async function caching<
+  S extends Store = Store,
+  T extends object = never,
+>(fac: Stores<S, T>, args?: CachingConfig<T>): Promise<Cache<S | MemoryStore>> {
```
BryanDonovan, line 64: I'd personally not use `fac`, but instead use the full word `factory`, etc.

package.json
```diff
@@ -25,18 +33,32 @@
...
-    "es6-promise": "^4.2.8",
-    "eslint": "7.7.0",
-    "jsdoc": "3.6.5",
-    "mocha": "8.1.1",
-    "nyc": "15.1.0",
-    "optimist": "0.6.1",
-    "sinon": "9.0.3"
+    "@commitlint/cli": "17.1.2",
+    "@commitlint/config-conventional": "17.1.0",
+    "@faker-js/faker": "7.5.0",
```
BryanDonovan, line 42: I contributed to Faker many years ago :)

src/caching.ts
BryanDonovan: I'd be hesitant to make the `caching` function (which is pretty much a constructor) be asynchronous.  Some users may want/need to instantiate their cache instances synchronously.  Those cases are probably just due to legacy code, however, so maybe that's not an issue.

BryanDonovan (approved):
Looks good to me.

src/caching.ts
botika: > I'd personally not use `fac`, but instead use the full word `factory`, etc.

Ok, I will change this variable names

BryanDonovan: Looks like the caching function being `async` is causing some issues (https://github.com/node-cache-manager/node-cache-manager/issues/210#issuecomment-1263920136).  Maybe the constructor should be synchronous and a separate method like `initialize()` or something can be async if needed.

salquier-appvizer:
Hi !
Can you add in the breaking changes that's the TTL for memory stores changes from seconds to milliseconds ?

botika:
> Hi ! Can you add in the breaking changes that's the TTL changes from seconds to milliseconds ?

You can do the PR yourself with those changes. Thanks.

salquier-appvizer:
Done !
https://github.com/node-cache-manager/node-cache-manager/pull/277

PR #256, opened 2022-11-02
fix: re populate higher priority caches when a key is found in lower ones

### LINKED ISSUE #253: v5 regression: wrap() no longer re-populates higher priority caches when a key is found in lower ones.
When using multi-level caches eg memory & redis, when a key is not found in the higher level cache (eg memory) but it is found in the lower one (redis), the key is no longer set in the higher levels. As a result subsequent requests for the same key will always have to reach redis and consume extra bandwidth.
In v4 wrap was re-populating the higher level caches, so subsequent requests within the ttl of the memory cache did not have to reach redis.

botika:
Fixes #253

### DIFF
--- a/src/multi-caching.ts
+++ b/src/multi-caching.ts
@@ -25,12 +25,23 @@ export function multiCaching<Caches extends Cache[]>(
     del: async (key) => {
       await Promise.all(caches.map((cache) => cache.del(key)));
     },
-    async wrap<T>(key: string, fn: () => Promise<T>): Promise<T> {
-      const value = await get<T>(key);
+    async wrap<T>(key: string, fn: () => Promise<T>, ttl?: Ttl): Promise<T> {
+      let value: T | undefined;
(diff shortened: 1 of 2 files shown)


src/multi-caching.ts
```diff
@@ -25,12 +25,23 @@ export function multiCaching<Caches extends Cache[]>(
...
+      }
       if (value === undefined) {
         const result = await fn();
-        await set<T>(key, result);
+        await set<T>(key, result, ttl);
         return result;
+      } else {
+        for (let j = 0; j < i; j++) {
+          await caches[j].set(key, value, ttl);
+        }
```
thgreasi, line 44: I guess that this could be done in parallel and we could avoid waiting for it to complete, like it was before afair.
```suggestion
for (let j = 0; j < i; j++) {
  caches[j].set(key, value, ttl).catch(() => {});
}
```

PR #282, opened 2022-12-01
fix: passing 0 to ttl argument does not work

nekocode:


nekocode:
Hi @botika , can you help review this PR?

src/stores/memory.ts
```diff
@@ -49,7 +49,7 @@ export function memoryStore(args?: MemoryConfig): MemoryStore {
     keys: async () => [...lruCache.keys()],
     mget: async (...args) => args.map((x) => lruCache.get(x)),
     async mset(args, ttl?) {
-      const opt = { ttl: ttl ? ttl : lruOpts.ttl } as const;
+      const opt = { ttl: ttl != null ? ttl : lruOpts.ttl } as const;
```
botika, line 52: It's `ttl ==! undefined`

src/stores/memory.ts
```diff
@@ -36,7 +36,7 @@ export function memoryStore(args?: MemoryConfig): MemoryStore {
   const lruOpts = {
     ...args,
     max: args?.max || 500,
-    ttl: args?.ttl ? args.ttl : 0,
+    ttl: args?.ttl != null ? args.ttl : 0,
```
botika, line 39: It's `ttl ==! undefined`

src/stores/memory.ts
```diff
@@ -69,7 +69,7 @@ export function memoryStore(args?: MemoryConfig): MemoryStore {
         throw new Error(`no cacheable value ${JSON.stringify(value)}`);
       if (shouldCloneBeforeSet) value = clone(value);
 
-      const ttl = opt ? opt : lruOpts.ttl;
+      const ttl = opt != null ? opt : lruOpts.ttl;
```
botika, line 72: It's `ttl ==! undefined`

test/stores/memory.test.ts
```diff
@@ -15,7 +15,7 @@ describe('memory store', () => {
     let store: MemoryStore;
 
     beforeEach(() => {
-      store = memoryStore();
+      store = memoryStore({ ttl: 1 });
```
botika, line 18: this can be broken other test. Recreate store in the test with this arguments

botika (changes requested):

src/stores/memory.ts
nekocode: `ttl != null` equals`ttl !== undefined && ttl !== null`

https://github.com/lodash/lodash/blob/master/isNil.js#L19

botika: But by typing it is impossible for it to be null.

botika:
Thanks!

PR #345, opened 2023-02-02
Introduced typescript alias type for ttl

mrcwbr:
Improved the typing of the TTL type to Milliseconds. this is best practice for typescript (https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) and helps developer to set the right unit  because of the IDE support. 

if your not sure if the ttl property is seconds or milliseconds you are now able to hover the property and get the unit

what do you think about this change?

### DIFF
--- a/src/caching.ts
+++ b/src/caching.ts
@@ -1,18 +1,18 @@
 import { MemoryCache, MemoryConfig, memoryStore } from './stores';
 
 export type Config = {
-  ttl?: Ttl;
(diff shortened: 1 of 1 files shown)


botika:
Ok, you can also export another one with the name `Ttl` so that there are no breaking changes. And the CI is broken :smile_cat:

botika:
Fixed in `5.1.5`

Issue #361, opened 2023-02-15
mget/mset does not exist when using multiCaching


deyhle:
This worked in < v5, but now `mget` and `mset` are undefined on a multiCaching instance. `store.mget` and `store.mset` (taken from [the example for `mget` in the readme](https://github.com/node-cache-manager/node-cache-manager#example-settinggetting-several-keys-with-mset-and-mget)) is also undefined because `store` is undefined on a multiCaching instance.

Is it still possible to use `mget` and `mset` with multiCaching instances?

The test mentions the methods, but does not test for them: https://github.com/node-cache-manager/node-cache-manager/blob/a1eebf72bd60188e37f32dfb0c1e03ac23045c00/test/multi-caching.test.ts#L41

botika:
Not at this time but it can be implemented.

botika:
Fixes in `5.2.0`. Thanks!

PR #366, opened 2023-02-21
feat: add mget(), mset(), mdel() to multiCache

deyhle:
Fix for #361. 

The implementation is similar to the functionality that existed in previous versions and is missing in v5. However, `mset` takes the same array-style arguments as the single-store version in v5 (as opposed to the pre-v5-style with key, value pair argument list).

It would be cool if you merged and released a new version to npm with it.


botika:
The PR has multiple things. I can't review it or approve run CI.

deyhle:
Sorry, the Pull Request updated itself with some other commits I needed to make my fork run as intermediate solution. I'll send another with just the fix

PR #367, opened 2023-02-22
feat: add mget(), mset(), mdel() to multiCache

deyhle:
Fix for #361. (clone of #366 but with a clean commit)

The implementation is similar to the functionality that existed in previous versions and is missing in v5. However, mset takes the same array-style arguments as the single-store version in v5 (as opposed to the pre-v5-style with key, value pair argument list).

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,29 @@ console.log(await multiCache.get('foo2'));
 
 // Delete from all caches
 await multiCache.del('foo2');
+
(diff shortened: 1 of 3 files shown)


botika:
LGTM, but the branch is outdate

botika (changes requested):
LGTM, but the branch is outdate

PR #480, opened 2023-05-21
sizeCalculation key,value swap

Mojtaba-NA:
type MemoryConfig in memoryStore is incorrect and could cause some confusion (it did for me a bit)

this is the type in node-lru-cache package
[https://github.com/isaacs/node-lru-cache/blob/main/src/index.ts#L219](https://github.com/isaacs/node-lru-cache/blob/main/src/index.ts#L219)

as you can see value is first annd key is second.


botika:
Can you put coverage in that part?

botika:
thanks!

Issue #532, opened 2023-07-11
Feature: wrap function, allow set ttl according to the return of fn.


peng-huang-ch:
Allows to dynamically set the value of ttl.
I want the set the use cache ttl according to the user role.

```ts
export interface User {
  role: string;
}

const ttl = (user: User) => {
  if (user?.role === 'admin') {
    return 5 * 60 * 1000;
  } else if (user?.role === 'user') {
    return 3 * 60 * 1000;
  }
  return 1 * 60 * 1000;
};

function getUer(id: string) {
  var cacheKey = "suer-id:" + id;
  // var ttl = ttl(user);
  return memoryCache.wrap(
    cacheKey,
    function (cacheCallback) {
      DB.find(id, cacheCallback);
    },
    { ttl: ttl }, // I want to set the ttl according to the user role.
  );
}
```
Maybe solution: 

https://github.com/node-cache-manager/node-cache-manager/blob/43593ce3389a225c99e35e642420649d6848c12d/src/caching.ts#L81C4-L89C7

```ts
wrap: async <T>(key: string, fn: () => Promise<T>, ttl?: Milliseconds | ((T) => Milliseconds)) => {
  const value = await store.get<T>(key);
  if (value === undefined) {
    const result = await fn();
    if (typeof ttl === 'function' ) {
      ttl = ttl(result);
    }
    await store.set<T>(key, result, ttl);
    return result;
  }
  return value;
}
```

botika:
Yes, it can be implemented if you need it. I didn't see it as necessary to do the type check, but if you have an implementation that needs this, go ahead with the PR. Thank you!

peng-huang-ch:
@botika Please review the [PR](https://github.com/node-cache-manager/node-cache-manager/pull/535), THX.

PR #535, opened 2023-07-14
feat: the ttl of wrap could be function

peng-huang-ch:
#532

Allows to set the `ttl` of the `wrap` function based on the return of fn.


botika (approved):

botika:
thx!

weyert:
Any plan to do a release to include this?

Issue #533, opened 2023-07-12
Feature: optional refresh strategy (stale-while-revalidate)


jonathanarezki:
Before version 5 there was the option for a "refreshThreshold" for the any store.
The ttl of the cached key was checked against the refreshThreshold.
If the ttl was less than the refreshThreshold, the value was returned immediately, the provided function was still triggered to refresh the cache.
The benefit would be a more conservative caching strategy.

botika:
I think, `ttl1 === tll0 - refreshThreshold`.

Or maybe I'm wrong but the `refresh threshold` is the same `time to live`. Every time you use the set method or its variants you can pass the ttl as a argument. 

Can you explain to me the difference between using refreshThreshold and using a smaller ttl?

jonathanarezki:
the refresh threshold allows for a new functionality.
Lets assume you define the ttl to 24h and the refresh threshold to 23h.
If any value in that store then gets checked if the remaining ttl is less than the refresh threshold.
If that is the case we still return the value directly but we run the function asynchronously and set the cache again.
That way its possible to refresh a cache before it completely runs out.

Essentialy you can reach a 100% cache coverage without complicated cron jobs.
The functionality got removed with version 5.0.0

botika:
With this we only get necessary calls to our database, since we don't know if this value is going to be used or it is going to be invalidated but we have to make the call at that time yes or yes. 
The functionality of a cache is to reduce the number of calls, not to increase it. 

Can you describe a situation where it is used, because this requires more logic in the set method. 

If I did another method with this functionality, would it work for you?

jonathanarezki:
I wouldn't put the logic into the set function.
It's better suited for the wrap function.

That logic is useful if the data isn't critical but called really often.
With this all the calls get returned asap even if the cache got stale, and only then get called to the db for a refresh.

Here is the PR of the old logic, that got removed with 5.0.0: https://github.com/node-cache-manager/node-cache-manager/pull/138

botika:
> I wouldn't put the logic into the set function. It's better suited for the wrap function.

`wrap` uses `set`

one use case?

jonathanarezki:
It's still not in the set function.
The wrap function is already a special case.
The refresh logic fits perfectly into the what the wrap function tries to solve.

A use case: we cash the result of a function with wrap that takes upwards of 10 seconds but the wrap function gets called a hundred times per second.
If the cache runs out we would slow down a lot.
[...]

PR #586, opened 2023-09-05
feat: optional refresh strategy (stale-while-revalidate) #533

jonathanarezki:
Added refresh threshold, to update stores before the ttl runs out.
Only the wrap function is affected.
If the the remaining ttl of a store is smaller than the refresh threshold,
the store gets set with the fn asynchronously.
The refresh threshold is optional and is not stored in the store it self, only in the cache instance as a config.


jaredwray:
@jonathanarezki thanks for doing this pull request. I will be reviewing this in the next 10 days and get back to you on this.

jaredwray:
Also, please make sure to update your branch to the latest.

jaredwray:
@jonathanarezki can you update the readme with this new functionality?

jonathanarezki:
I updated the readme.

jaredwray (approved):

jaredwray:
@jonathanarezki - thanks and since this is not a non breaking change I will add it to the next release which should go out in the next couple weeks.

jonathanarezki:
Thank you really much. I didn't expect it to go that smooth.

PR #598, opened 2023-09-29
feat: support sync cache creation

denkan:

Add non-async `createCache()` that `caching()` uses to set up methods (keep current functionality intact).

Having `createCache()` as addition to `caching()` allows devs to create cache instances synchronously, and i.e use as default params in methods:

```ts
/** Awesome service that will use memory store as default cache, if nothing else specified */ 
class MyItemService {
   constructor(public cache = createCache(memoryStore())) { }
}
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: ['16', '18', '20']
+        node-version: ['18', '20']
 
     steps:
       - uses: actions/checkout@v3
--- a/README.md
+++ b/README.md
@@ -84,6 +84,32 @@ You can use your own custom store by creating one with the same API as the built
 - [Example Custom Store redis](https://github.com/node-cache-manager/node-cache-manager-redis-yet)
 - [Example Custom Store ioredis](https://github.com/node-cache-manager/node-cache-manager-ioredis-yet)
 
+#### Create single cache store synchronously
+
+As `caching()` requires async functionality to resolve some stores, this is not well-suited to use for default function/constructor parameters etc.
+
+If you need to create a cache store synchronously, you can instead use `createCache()`:
+
+```typescript
+import { createCache, memoryStore } from 'node-cache-manager';
+
+// Create memory cache synchronously
+const memoryCache = createCache(memoryStore(), {
+  max: 100,
+  ttl: 10 * 1000 /*milliseconds*/,
+});
+
+// Default parameter in function
+function myService(cache = createCache(memoryStore())) {}
+
+// Default parameter in class constructor
+const DEFAULT_CACHE = createCache(memoryStore(), { ttl: 60 * 1000 });
+// ...
+class MyService {
+  constructor(private cache = DEFAULT_CACHE) {}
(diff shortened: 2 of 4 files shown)


douglascayers (approved):
I'm a consumer of this lib like you, and for what it's worth, your changes look good to me, thanks!

We recently migrated from v4 to v5 and were a bit surprised that the method to create a cache became async. Your change will make those migrations easier, plus the benefits of when needing to instantiate the cache in a sync environment (e.g. constructors)

jaredwray:
@denkan - can you add in an update to the readme on how to use this?

jaredwray:
@denkan - thanks so much for doing this. One thing I was thinking is that we might want to change it to be `createMemoryCache` as a simple helper function. The reason is that it could be confusing to do this function as `sync` when if they put in an `async` store it will cause issues. 

You could even simplify and not require them to put in the in-memory store if you wanted. 


Can you change that over and update the branch to latest? If you are good with these changes and update it we can check this in. 🎉

alfonsograziano:
I'd love to use this feature as well!

jaredwray:
@casey-chow - we are ready to merge this. Can you do the following:
- update your branch to the latest and resolve conflicts
- make sure you are at 100% test coverage
- make sure documentation with how to use this is updated

🍻

jaredwray:
@casey-chow are you able to update the branch?

denkan-ikea:
@jaredwray and others,
Sorry I dropped the ball on this one (as I mostly use another github account for daily work nowadays).

✅  I've rebased and resolved conflicts.


> @denkan - thanks so much for doing this. One thing I was thinking is that we might want to change it to be `createMemoryCache` as a simple helper function. The reason is that it could be confusing to do this function as `sync` when if they put in an `async` store it will cause issues.
> 
> You could even simplify and not require them to put in the in-memory store if you wanted.

I don't really want to do this, as it's not actually limited to memoryStore? I could easily define custom store that doesn't require async initialization.

And I think it's better to let the user of the lib define a shorthand if needed:
```js
export const defaultCache = () => createCache(memoryStore({ /* my options */ })); 

// later
class MyItemService {
   constructor(public cache = defaultCache()) { }
}
```


Let me know your thoughts or if it's ready to merge 🍻

denkan:
@jaredwray and others,
Sorry I dropped the ball on this one (as I mostly use another github account for daily work nowadays).

✅  I've rebased and resolved conflicts.


> @denkan - thanks so much for doing this. One thing I was thinking is that we might want to change it to be `createMemoryCache` as a simple helper function. The reason is that it could be confusing to do this function as `sync` when if they put in an `async` store it will cause issues.
> 
> You could even simplify and not require them to put in the in-memory store if you wanted.

I don't really want to do this, as it's not actually limited to memoryStore? I could easily define custom store that doesn't require async initialization.

And I think it's better to let the user of the lib define a shorthand if needed:
```js
export const defaultCache = () => createCache(memoryStore({ /* my options */ })); 

// later
class MyItemService {
   constructor(public cache = defaultCache()) { }
}
```


Let me know your thoughts or if it's ready to merge 🍻

PR #599, opened 2023-10-01
fix(caching): coalesce concurrent requests to `wrap` function

### LINKED ISSUE #417: callbackFiller removed in 5.0.0
I would like to know if there was a particular reason why this feature was removed? I might be misunderstanding the current behavior of `.wrap()` in v5.0.0, but by looking at the code, it seems like there isn't any mechanism left to coalesce simultaneous cache wrap for the same key, unlike before with `callbackFiller`. 

If I'm not mistaken and there isn't indeed a mechanism currently that does that, would you be open of a PR that re-introduce a `callbackFiller`-like feature? We were about to update before noticing this, which would really affect our application with how we've expected node-cache-manager to work, locking us out of the v5.0.0 with the await/async and Typescript support.

douglascayers:

Fixes #417

In the `v4.x` version, the `caching.wrap()` function would coalesce multiple concurrent requests to minimize how many times the generator function was invoked. This was a helpful optimization to reduce load on downstream systems like databases and APIs.

This pull request adds that cache miss relief logic back by using the [promise-coalesce](https://github.com/douglascayers/promise-coalesce) module, which is said logic in a generic, reusable function.

For tests, the `#417` issue test demonstrates the undesired behavior. The updated `wrap()` tests demonstrate the desired behavior by coalescing requests.

Thanks

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
   "license": "MIT",
   "dependencies": {
     "lodash.clonedeep": "^4.5.0",
-    "lru-cache": "^10.0.1"
+    "lru-cache": "^10.0.1",
+    "promise-coalesce": "^1.1.1"
   },
   "devDependencies": {
     "@faker-js/faker": "8.1.0",
--- a/src/caching.ts
+++ b/src/caching.ts
@@ -1,3 +1,4 @@
+import { coalesceAsync } from 'promise-coalesce';
 import { MemoryCache, MemoryConfig, memoryStore } from './stores';
 
 export type Config = {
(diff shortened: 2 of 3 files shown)


jaredwray:
Hi @douglascayers - thanks for putting this together. A couple of questions around this:

- With this change is there anything breaking or should everything work as it with this added functionality?
- Since you added this module, will you be able to support and maintain it? If not, I would prefer adding the code directly into this project. 
- Can you update the readme with an example of how this works?

douglascayers:
@jaredwray Thanks for the follow up. I've rebased this PR and have answers to your questions below. Thanks

> * With this change is there anything breaking or should everything work as it with this added functionality?

This is a drop-in patch update. No breaking changes. Users don't have to make any changes to benefit from this change.

The functional difference is that cache manager's `wrap` function is more efficient when multiple concurrent requests occur for the same cache key. 

When there's a cache miss, instead of calling the user's function to fetch a value to cache once per request (which can exhaust downstream API rate limits, be slower, etc), one promise goes off to do the work and each of the other concurrent requests receive a promise that _promises_ to resolve to the same value as the first promise that's actually executing the user's function.


> * Since you added this module, will you be able to support and maintain it? If not, I would prefer adding the code directly into this project.

Yes, I intend to support it. I see value in the coalescing logic as a standalone helper library, which is agnostic to caching though that is the main driver for the solution.

With that said, I'm not opposed to adding the code directly to this project if you all want to maintain it going forward.

> * Can you update the readme with an example of how this works?

The README for cache manager doesn't need to change since users don't have to do anything differently than they do now. If this PR is merged, then users will notice an efficiency boost when concurrent requests call `wrap` function for the same cache-key.

This PR includes new tests to demonstrate the problem when `wrap` is called concurrently with and without coalescing logic.

I also have an example on the [promise-coalesce](https://github.com/douglascayers/promise-coalesce?tab=readme-ov-file#example) project's README, too.


Let me know if you have any other questions or concerns, thanks

jaredwray:
@douglascayers - thanks so much for all the work on this and we will be working to get this deployed in the next couple weeks.

douglascayers:
Fantastic, thanks!

PR #604, opened 2023-10-08
fix(refreshThreshold): don't run if ttl is -1

mihirgupta0900:

**What kind of change does this PR introduce?** (Bug fix)

In stores like redis, if no `ttl` value is set, `store.ttl(key)` returns `-1`. This causes an issue with the `refreshThreshold` feature, where the comparison mentioned below will always lead to running the function

https://github.com/node-cache-manager/node-cache-manager/blob/2eca5a0c6c905fe706ed9eaca9152dedd512ef7e/src/caching.ts#L92

I've added an additional check to check for `-1` value, and not run the function again.

Note: Wasn't sure where to add test for this, please let me know if I need to. I have tested this manually with my work project

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# node-cache-manager 
+# node-cache-manager
+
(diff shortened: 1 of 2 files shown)


jaredwray:
@mihirgupta0900 - thanks for this fix. This should be released with our next update which will be in 2-3 weeks. 🍻

PR #620, opened 2023-11-23
619: Ensure that background refresh only calls fn once.

ricall:
Bug fix for https://github.com/node-cache-manager/node-cache-manager/issues/619


jaredwray:
@ricall - this looks to still be failing. Any update on this?

jaredwray:
@ricall - this is fixed now. Thanks!

PR #637, opened 2024-01-30
Ability to catch promise rejections from background refreshes

Kauhsa:
This PR introduces `onBackgroundRefreshError` configuration option, which makes it possible to handle promise rejections originating from background refreshes.

As it stands now, depending on your Node version, process might crash. `process.on('uncaughtException', ...)` fixes that, but we missed a bit more... targeted approach to handling these errors, so here is a PR.

### DIFF
--- a/src/caching.ts
+++ b/src/caching.ts
@@ -5,6 +5,7 @@ export type Config = {
   ttl?: Milliseconds;
   refreshThreshold?: Milliseconds;
   isCacheable?: (val: unknown) => boolean;
+  onBackgroundRefreshError?: (error: unknown) => void;
 };
 
 export type Milliseconds = number;
@@ -114,9 +115,15 @@ export function createCache<S extends Store, C extends Config>(
           const cacheTTL = typeof ttl === 'function' ? ttl(value) : ttl;
           const remainingTtl = await store.ttl(key);
           if (remainingTtl !== -1 && remainingTtl < refreshThresholdConfig) {
-            coalesceAsync(`+++${key}`, fn).then((result) =>
-              store.set<T>(key, result, cacheTTL),
-            );
+            coalesceAsync(`+++${key}`, fn)
+              .then((result) => store.set<T>(key, result, cacheTTL))
+              .catch((error) => {
+                if (args?.onBackgroundRefreshError) {
+                  args.onBackgroundRefreshError(error);
+                } else {
+                  return Promise.reject(error);
+                }
+              });
           }
         }
         return value;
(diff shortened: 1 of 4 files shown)


src/caching.ts
```diff
@@ -114,9 +115,15 @@ export function createCache<S extends Store, C extends Config>(
...
-            coalesceAsync(`+++${key}`, fn).then((result) =>
-              store.set<T>(key, result, cacheTTL),
-            );
+            coalesceAsync(`+++${key}`, fn)
+              .then((result) => store.set<T>(key, result, cacheTTL))
+              .catch((error) => {
+                if (args?.onBackgroundRefreshError) {
+                  args.onBackgroundRefreshError(error);
+                } else {
+                  return Promise.reject(error);
```
Kauhsa, line 124: This should behave as before if `onBackgroundRefreshError` is not set. However, should this error be anyway wrapped to some custom error (`BackgroundRefreshError`...?) to make it easier to figure out where the unhandled promise is originating from?

test/utils.ts
```diff
@@ -1 +1,25 @@
...
+  /* restore previous listeners */
+  return () => {
+    uncaughtExceptionListeners.forEach((listener) =>
+      process.addListener('uncaughtException', listener),
+    );
+    unhandledRejectionListeners.forEach((listener) =>
+      process.addListener('unhandledRejection', listener),
+    );
+  };
+};
```
Kauhsa, line 25: Vitest injects its own listeners, which made it a bit difficult to test whether `process.on('uncaughtException')` works as expected. This deals with that.

test/utils.ts
```diff
@@ -1 +1,25 @@
...
+  /* restore listeners */
+  return () => {
+    uncaughtExceptionListeners.forEach((listener) =>
+      process.addListener('uncaughtException', listener),
+    );
+    unhandledRejectionListeners.forEach((listener) =>
+      process.addListener('unhandledRejection', listener),
+    );
+  };
+};
```
Kauhsa, line 25: Vitest injects its own listeners, which made it a bit difficult to test whether process.on('uncaughtException') works as expected. This deals with that.

jaredwray:
@Kauhsa can we get an update on this?

Kauhsa:
> @Kauhsa can we get an update on this?

Sure. Rebased.

jaredwray:
@Kauhsa - thanks so much for adding this in and updating the readme. We will get this published in the next couple weeks.

PR #646, opened 2024-03-03
Add onCacheError option

slukes:

- [x]  Followed the [Contributing](https://github.com/node-cache-manager/node-cache-manager/blob/main/CONTRIBUTING.md) guidelines.

This PR addresses:

https://github.com/node-cache-manager/node-cache-manager/issues/643
https://github.com/node-cache-manager/node-cache-manager/issues/588
https://github.com/node-cache-manager/node-cache-manager/issues/51

By re-adding an option that [existed](https://github.com/node-cache-manager/node-cache-manager/pull/4/files) in caching.js to 'ignore' cache errors but was [removed](https://github.com/node-cache-manager/node-cache-manager/pull/203) during the migration to typescript / promises.

This is useful to applications that are using a distant cache store, to enable to application to continue to work, even when the cache backend is down.


jaredwray:
@slukes - I would rather not have another config on this and since it is cache should we just add in EventEmitter3 and emit an error when it goes to a catch?

slukes:
> @slukes - I would rather not have another config on this and since it is cache should we just add in EventEmitter3 and emit an error when it goes to a catch?

Hello @jaredwray I'm not sure how this is going to work!  I've created a ticket on our team backlog to dig into this further in the next couple of weeks

jaredwray:
@slukes - what I am saying is that instead of doing an `ignoreErrors` property in the cache manager what we can do is enable you to do this instead:

```
import { caching } from 'cache-manager';

const cache = await caching('memory', {
  max: 100,
  ttl: 10 * 1000 /*milliseconds*/,
});

cache.on('error', (error) = > {
     console.log(error);
});
```

This would use events to emit an error and by default we would make all methods catch errors and not throw. This gives the resiliency shown in `keyv` that many love. 

Would that work for you?

src/multi-caching.ts
```diff
@@ -26,20 +30,28 @@ export function multiCaching<Caches extends Cache[]>(
 		data: T,
 		ttl?: Milliseconds | undefined,
 	) => {
-		await Promise.all(caches.map(async cache => cache.set(key, data, ttl)));
+		try {
+      await Promise.all(caches.map(async cache => cache.set(key, data, ttl)));
```
slukes, line 34: @AdrienDoctolib I think we should talk about this - might be better to use allSettled

test/caching.test.ts
```diff
@@ -319,6 +319,40 @@ describe('caching', () => {
 				});
 			}
 		});
+
+		describe('error handling on wrap', () => {
+			it('receives an error when store.get() fails', async () => {
```
slukes, line 324:
```suggestion
			it('emits and error event when store.get() fails', async () => {
```

test/caching.test.ts
```diff
@@ -319,6 +319,40 @@ describe('caching', () => {
...
+				cache.store.get = vi.fn().mockRejectedValue(error);
+
+				let errorMessage;
+				cache.on('error', error => {
+					errorMessage = error;
+				});
+
+				await cache.wrap(key, function_);
+
+				expect(errorMessage).not.toBeUndefined();
```
slukes, line 337:
```suggestion
				expect(errorMessage).not.toBeUndefined();
				expect(function_).toHaveBeenCalled();
```

test/caching.test.ts
```diff
@@ -319,6 +319,40 @@ describe('caching', () => {
...
+			});
+
+			it('receives an error when store.set() fails', async () => {
```
slukes, line 340:
```suggestion
			it('emits an error event when store.set() fails', async () => {
```

test/multi-caching.test.ts
```diff
@@ -233,6 +248,30 @@ describe('multiCaching', () => {
 				await multiCache.set(key, value);
 				await expect(multiCache.get(key)).resolves.toBeUndefined();
 			});
+
+			it('should receive error event on get failure', async () => {
```
slukes, line 252:
```suggestion
			it('emits an error event when store.get() fails', async () => {
```

slukes:
@jaredwray thanks to the work of @AdrienDoctolib we now have something that is working how you suggested.  Please let us know your feedback. 

We have tested in our application and all is well 👍

jaredwray:
@AdrienDoctolib and @slukes looks like we are getting an error on compile time:

Screenshot 2024-04-03 at 11 09 29 AM

slukes:
> @AdrienDoctolib and @slukes looks like we are getting an error on compile time:

Sorry about that - fixed!

PR #714, opened 2024-07-02
non-blocking set for wrap function

corradodellorusso:
- [x] Followed the [Contributing](https://github.com/jaredwray/cache-manager/blob/main/CONTRIBUTING.md) guidelines.

Introducing non-blocking set for wrap operations. 
In V4 when the wrap operation was called, results were returned before backfilling the cache(s).
This changed in V5 and cache(s) are backfilled before returning results, leading to slower throughput. 

This PR adds the option to restore the behaviour of V4.

### DIFF
--- a/packages/cache-manager/README.md
+++ b/packages/cache-manager/README.md
@@ -232,6 +232,17 @@ multicache.on('error', (error) => {
 });
 ```
 
+## Using non-blocking set with wrap
+By default, when using `wrap` the value is set in the cache before the function returns. 
+While this behaviour can prevent additional calls to downstream resources, it can also slow down the response time.
+This can be changed by setting the `nonBlockingSet` option to `true`. 
+Doing will make the function return before the value is set in the cache.
+The setting applies to both single and multi caches. 
+
+```typescript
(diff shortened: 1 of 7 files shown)


jaredwray:
@corradodellorusso  would we want to extend this to the entire library or just wrap?

corradodellorusso:
> @corradodellorusso would we want to extend this to the entire library or just wrap?

@jaredwray as far as I understand `wrap` is the only operation that involves multiple different operations in its implementation, for the others operation is sufficient to omit `await` to obtain the same result.

jaredwray:
@corradodellorusso - we have added nonBlocking as a core but have not implemented it in wrap yet. Do you still need that?

corradodellorusso:
Hello Jared, I’m not sure I got the question.In the version we currently use (5.7.6) we implemented the nonBlockingSet as a parameter for wrap on the single cache.Does this answer your question?Inviato da iPhoneIl giorno 25 set 2024, alle ore 18:05, Jared Wray ***@***.***> ha scritto:﻿
@corradodellorusso - we have added nonBlocking as a core but have not implemented it in wrap yet. Do you still need that?

jaredwray:
Thanks. We added non blocking functionality across the rest of the service also with v6

Issue #742, opened 2024-08-05
 cache.wrap is shared between cache instances


kentandersen:
**Describe the bug**
When invoking `cache.wrap()` simultaneously with the same key on two different instances, the result from the first method is returned for both invocations. This has major implications when the two caches store different data, or data-structures, which can result in wrong result or worst case reference error.

The culprit is `promise-coalesce` which stores all invocations in a global Map, regardless of which cache the invocation came from.

A fix can be include a per cache prefix to all coalesceAsync invocations.

**How To Reproduce (best to provide workable code or tests!)**
```
import { createCache, memoryStore } from 'cache-manager'

const cache1 = createCache(
    memoryStore({
        max: 50,
        ttl: 60000,
    }),
)

const cache2 = createCache(
    memoryStore({
        max: 50,
        ttl: 60000,
    }),
)

const [result1, result2] = await Promise.all([
    cache1.wrap('THE_SAME_KEY', () => Promise.resolve({ value: 1 })),
    cache2.wrap('THE_SAME_KEY', () => Promise.resolve({ value: 2 })),
])

console.log(result1) // { value: 1 }, correct
console.log(result2) // { value: 1 }, expected { value: 2 }
```

jaredwray:
This has been released

PR #743, opened 2024-08-05
fixing collision when invoking wrap on multiple cache instances with …

kentandersen:
…same key simultaneously (#742)

- [x] Followed the [Contributing](https://github.com/jaredwray/cache-manager/blob/main/CONTRIBUTING.md) guidelines.

Fix for issue #742; when invoking `cache.wrap()` simultaneously with the same key on two different instances, the result from the first method is returned for both invocations


jaredwray:
@kentandersen - I haven't had time to look into this but wanted to make sure that we are not affecting the key itself being stored such as in redis. Is this only for the coalescAsync key id?

kentandersen:
The only keys affected by this change is the ones supplied to coalescAsync. I even named the prefix variable `coalescePrefix` to make it clear that it should never be used with any other key

PR #887, opened 2024-11-08
file-entry-cache - fix recreating cache from a file

### LINKED ISSUE #886: File-entry-cache does not recognize file changes when loading cache from a file
**Describe the bug**
When file-entry-cache creates a new instance from a file, and a "cached" file had changed before cache instance was created, it reports that the "cached" file did not change.

1. There is a cache file containing metadata of file X
1. File X changes
1. file-entry-cache creates a new cache instance from the cache file
1. file-entry-cache `analyzeFiles` and `getUpdatedFiles` do not report file X as changed

This is a regression in version 10.0.0 - according to npm's versioning. I don't know how `cacheable` date based versioning relates to npm's semver.

**How To Reproduce**
Codesandbox example: https://codesandbox.io/p/devbox/c3rx6r
From the terminal you can run:
- `yarn works` -  works as expected
- `yarn fails1` - does not detect file changes
- `yarn fails2` - does not detect file changes.
If in `package.json` you switch to `file-entry-cache` version 9.1.0 and run `yarn` to reinstall packages, then `fail1` and `fail2` scripts detect file changes.

I will also provide unit tests in a pull request.

lkinasiewicz:
Fixes #886

### DIFF
--- a/packages/file-entry-cache/src/index.ts
+++ b/packages/file-entry-cache/src/index.ts
@@ -58,7 +58,6 @@ export function create(cacheId: string, cacheDirectory?: string, useCheckSum?: b
 		const cachePath = `${cacheDirectory}/${cacheId}`;
 		if (fs.existsSync(cachePath)) {
 			fileEntryCache.cache = createFlatCacheFile(cachePath, options.cache);
-			fileEntryCache.reconcile();
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


jaredwray:
@lkinasiewicz - this should be released in the next 2-3 weeks. Will that work?

lkinasiewicz:
Thanks for quick response! No hurry, I will keep using version 9.1.0 for a while. 🙂

PR #899, opened 2024-11-16
cache-manager - adding disconnect to cache

TimothyIp:

Adds the ability to disconnect from the relevant stores inside the cache

### DIFF
--- a/packages/cache-manager/README.md
+++ b/packages/cache-manager/README.md
@@ -336,7 +336,7 @@ See unit tests in [`test/wrap.test.ts`](./test/wrap.test.ts) for more informatio
 
 `disconnect(key): Promise<void>`
 
-Will disconnect from the relevant store(s).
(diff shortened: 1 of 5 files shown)


packages/cache-manager/README.md
```diff
@@ -332,6 +332,18 @@ await cache.wrap('error', () => {
 
 See unit tests in [`test/wrap.test.ts`](./test/wrap.test.ts) for more information.
 
+## disconnect
+
+`disconnect(key): Promise<void>`
+
+Will disconnect from the relevant store(s).
```
TimothyIp, line 339: Let me know if we should add a small disclaimer about using disconnect only when needed. Depending on the store, for example, redis. It is best practice to re-use connections if possible

jaredwray: That would be good to add in launguage about understanding the Keyv driver you are using and when you should use disconnect.

jaredwray: I added it.

jaredwray:
@TimothyIp - thanks so much and will get this rolled out in the next couple weeks.

TimothyIp:
This is awesome! Thanks!

PR #919, opened 2024-12-03
cacheable - unref check interval to allow node exit

mothershipper:
When a check interval is passed to the cache, cachable starts an interval that prevents node from exiting unless it gets cleared. There is a public method to stop the interval, but it's flagged as internal use only.

Given this, I've opted to unref the interval to allow the node process to exit cleanly.

The other alternative would be to add a public close/disconnect API method -- the Keyv interface seems to account for this case -- however, it does make the API more cumbersome for the end-user.


jaredwray:
@mothershipper - thanks so much and really appreciate it. This should work for most cases.

PR #943, opened 2024-12-17
preserve args type when wrapping

ebceu4:
Preserve args type when wrapping


jaredwray:
@ebceu4 - thanks for the PR. This should be released in the next couple weeks.

PR #965, opened 2025-01-02
on wrap with refreshThreshold - re-evaluate ttl function

yaronyam:
`wrap` function: re-evaluate ttl function on fresh value when triggered by `refreshThreshold`. rather then use ttl evaluated for cached value.


jaredwray:
@yaronyam thanks for fixing this. really appreciate it! Will most likely release this fix in a couple of weeks.

PR #967, opened 2025-01-03
bugfix wrap function: nested wrap call on other cache instance - results in deadlock

yaronyam:
`wrap` function: nested `wrap` call on other cache instance with same key - results in deadlock, where `wrap` function call hangs indefinitely

### DIFF
--- a/packages/cache-manager/README.md
+++ b/packages/cache-manager/README.md
@@ -181,6 +181,10 @@ const cache = createCache({ stores: [keyv] });
     * `clear` - will not wait for all stores to finish.
     * `wrap` - will do the same as `get` and `set` (return the first value found and not wait for all stores to finish).
 
+- **cacheId**?: string - Defaults to random string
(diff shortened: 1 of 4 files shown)


jaredwray:
@yaronyam - can you resolve the conflicts from the previous merge so I can review?

yaronyam:
sure. updated

jaredwray:
@yaronyam - thanks for your work on this. I make this an enhancement as I would like to provide the `cacheId` option on the options.

yaronyam:
cool 👍

jaredwray:
> cool 👍

Let me know when you get those changes in and I will get this merged.

yaronyam:
updated with `cacheId` as an optional cache option, defaults to random string.
added `Cache#cacheId()` method to return `cacheId`.

jaredwray (dismissed):

PR #975, opened 2025-01-12
cache-manager: wrap() method - support refreshThreshold argument as Function

yaronyam:
cache-manager: `wrap()` method - support `refreshThreshold` argument as number or Function  of `(value:T) => number`. Similar to how `ttl` is supported.

#### Motivation
`refreshThreshold` is ideally defined in proportion to `ttl` - which is supported also as a Function.
One popular use case for example: caching jwt tokens - cache `ttl` should be set according to returned jwt expiration. Likewise, `refreshThreshold` needs to be set proportionally. Especially when different clients change their jwt token lifetime dynamically.


jaredwray:
@yaronyam - can you get the lint errors fixed?

yaronyam:
Yes, fixed.

jaredwray:
@yaronyam - thanks. This should be live in a week or so.

Issue #997, opened 2025-01-28
Extend cache-manager wrap function to return raw data


vz-tl:
**Is your feature request related to a problem? Please describe.**
I'm using cache-manager (v6) as part of an Express app by utilizing `wrap` function to retrieve data from a cache (Redis in this case). After the data is returned, a cache-control header will be added to the response with `max-age` set to the remaining cache lifetime based on the cache entitie's expiration timestamp.

Using the wrap method, this requires to subsequently get the expiration value via `cache.ttl(key)`, even tough wrap already retrieves the data from the cache via `cache.get(key, { raw: true })`, which already returns the expiry. So this is basically a superfluous request to the cache instance I've to do just for obtaining the expiration, which is already known and could simply be returned by the `wrap` function.

**Describe the solution you'd like**
A possible solution could be to extend the `wrap` method with an optional `raw` parameter, for returning the raw data including the expiration timestamp.

I've monkey-patched cache-manager to be able proceed with my task, however it would be great if my proposal could be considered for being added to this awesome package. Below find the monkey-patch I added:

```patch
diff --git a/node_modules/cache-manager/dist/index.d.ts b/node_modules/cache-manager/dist/index.d.ts
index 5568105..3f223a3 100644
--- a/node_modules/cache-manager/dist/index.d.ts
+++ b/node_modules/cache-manager/dist/index.d.ts
@@ -57,7 +57,8 @@ type Cache = {
     del: (key: string) => Promise<boolean>;
     mdel: (keys: string[]) => Promise<boolean>;
     clear: () => Promise<boolean>;
-    wrap: <T>(key: string, fnc: () => T | Promise<T>, ttl?: number | ((value: T) => number), refreshThreshold?: number | ((value: T) => number)) => Promise<T>;
+    wrap<T>(key: string, fnc: () => T | Promise<T>, ttl?: number | ((value: T) => number), refreshThreshold?: number | ((value: T) => number)): Promise<T>;
+    wrap<T>(key: string, fnc: () => T | Promise<T>, ttl?: number | ((value: T) => number), refreshThreshold?: number | ((value: T) => number), raw: true): Promise<{ value: T, expires: number}>;
     on: <E extends keyof Events>(event: E, listener: Events[E]) => EventEmitter;
     off: <E extends keyof Events>(event: E, listener: Events[E]) => EventEmitter;
     disconnect: () => Promise<undefined>;
[...]

PR #998, opened 2025-01-30
cache-manager:wrap() method - support options parameter and raw return data

vz-tl:
cache-manager: wrap() method:
-  add support for an options object parameter instead of the separate `ttl` and `refreshThreshold` params.
- additionally a new parameter is added to the options object, that will change the method's return value to raw stored data including expiration timestamp.

Closes https://github.com/jaredwray/cacheable/issues/997


vz-tl:
@jaredwray , there seems to be an issue with the checks. Could you help out?

jaredwray:
It looks like it is failing on multi-layer tests:

```
× wrap with multi-layer stores > should refresh according to refreshThreshold 513ms
       → expected +0 to deeply equal 1
     × wrap with multi-layer stores > should respect refreshAllStores 552ms
       → expected +0 to deeply equal 1
```

vz-tl:
should be fixed now.

PR #1005, opened 2025-02-12
cache-manager - fix: return types of mget to handle Array<T | null>

vinnymac:
- [ ] Tests for the changes have been added (for bug fixes/features) with 100% code coverage.

This is a typescript bug fix. Currently the types have two bugs. The type is cast to be an array of a single element `T`. In reality the type can be an array with multiple values of element `T` or `null`.

### DIFF
--- a/packages/cache-manager/src/index.ts
+++ b/packages/cache-manager/src/index.ts
@@ -121,14 +121,14 @@ export const createCache = (options?: CreateCacheOptions): Cache => {
 	};
 
 	const mget = async <T>(keys: string[]) => {
-		const result = [];
+		const result: (T | null)[] = [];
 
 		for (const key of keys) {
 			const data = await get<T>(key);
 			result.push(data);
 		}
 
-		return result as [T];
+		return result;
 	};
 
 	// eslint-disable-next-line @typescript-eslint/ban-types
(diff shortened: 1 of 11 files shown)


jaredwray:
Hi @vinnymac - thanks for looking into this. I would suggest that we moved to `undefined` as for most cases, returning undefined for a missing cache entry is preferable because it aligns with standard JavaScript behavior and avoids confusion when distinguishing between "not set" and "explicitly set to null."

Also, looks like there are some errors when doing the build that you will need to update and fix.

jaredwray:
@vinnymac - so sorry on this as I didnt remember that all the other functions were `null`. can you roll it back to be null so we dont create a breaking change?

jaredwray:
My fault for not looking on this and will get it fixed.

vinnymac:
I'm okay with the null return personally, why did you choose to rollback from undefined though?

jaredwray:
otherwise it is a breaking change and we try to avoid that. Realistically we do one major version per year.

vinnymac:
Makes sense to me! Perhaps if you have a list of things for the next major we can add this to it?

jaredwray:
Yep. Already added it. I think we will most like release the next major version of cache-manager in Q3. It will most likely have a new unified wrap function that we have been working on.

PR #1045, opened 2025-03-19
TTL Propagation Bug Fixes

TRiLeZ:

TTLs are not propagated when utilizing multiple cacheable instances with a shared secondary store.

This PR starts with adding test cases that define the expected behavior of TTL propagation. For the added cases, we use two cacheable instances that utilize a shared secondary store, writing to the first and then reading from the second. It helps to consider the use-case where we have two processes using the same Redis database as the secondary store.

Test case status: 7/7 passing.

### DIFF
--- a/packages/cacheable/src/index.ts
+++ b/packages/cacheable/src/index.ts
@@ -301,8 +301,22 @@ export class Cacheable extends Hookified {
 				const rawResult = await this._secondary.get(key, {raw: true});
 				if (rawResult) {
 					result = rawResult.value as T;
-					const finalTtl = rawResult.expires ?? undefined;
-					await this._primary.set(key, result, finalTtl);
+
+					let finalTtl;
+					let expired = false;
+					if (rawResult.expires) {
+						const now = Date.now();
+						finalTtl = rawResult.expires - now;
+						if (finalTtl <= 0) {
+							expired = true;
+						}
+					}
+
+					if (expired) {
+						result = undefined;
+					} else {
+						await this._primary.set(key, result, finalTtl);
+					}
 				}
 			}
 
--- a/packages/cacheable/test/index.test.ts
+++ b/packages/cacheable/test/index.test.ts
@@ -317,6 +317,88 @@ describe('cacheable set and get with ttl', async () => {
 		const result = await cacheable.get('key');
 		expect(result).toEqual('value');
 	});
+
+	describe('ttl propagation using a shared secondary store', async () => {
+		test('should get a value from the secondary store and respect its ttl when setting the value in the primary store (item specific ttl)', async () => {
+			const instance1Primary = new Keyv();
+			const instance2Primary = new Keyv();
+
+			const sharedSecondary = new Keyv();
+
+			const instance1 = new Cacheable({primary: instance1Primary, secondary: sharedSecondary});
+			const instance2 = new Cacheable({primary: instance2Primary, secondary: sharedSecondary});
+
+			// Set the value in the first instance
+			await instance1.set('key', 'value', 500);
+
+			await sleep(100);
+
+			// Get the value in the second instance
+			const result = await instance2.get('key');
+			expect(result).toEqual('value');
+
+			// Wait for the value to expire
+			await sleep(700);
+
+			// Get the value in the second instance (it should be expired)
+			const result2 = await instance2.get('key');
+			expect(result2, 'result should have expired').toBeUndefined();
+		});
+
+		test('should get a value from the secondary store and respect its ttl when setting the value in the primary store (default ttl when setting)', async () => {
+			const instance1Primary = new Keyv();
+			const instance2Primary = new Keyv();
+
+			const sharedSecondary = new Keyv();
+
+			const instance1 = new Cacheable({primary: instance1Primary, secondary: sharedSecondary, ttl: 500});
+			const instance2 = new Cacheable({primary: instance2Primary, secondary: sharedSecondary});
+
+			// Set the value in the first instance
+			await instance1.set('key', 'value');
+
+			await sleep(100);
+
+			// Get the value in the second instance
+			const result = await instance2.get('key');
+			expect(result).toEqual('value');
+
+			// Wait for the value to expire
+			await sleep(700);
+
+			// Get the value in the second instance (it should be expired)
+			const result2 = await instance2.get('key');
+			expect(result2, 'result should have expired').toBeUndefined();
+		});
+
+		test('should get a value from the secondary store and respect its ttl when setting the value in the primary store (default ttl when setting in the first instance, despite alternative ttl when getting in the second instance)', async () => {
+			const instance1Primary = new Keyv();
+			const instance2Primary = new Keyv();
+
+			const sharedSecondary = new Keyv();
+
+			const instance1 = new Cacheable({primary: instance1Primary, secondary: sharedSecondary, ttl: 500});
+			const instance2 = new Cacheable({primary: instance2Primary, secondary: sharedSecondary, ttl: 2000});
+
+			// Set the value in the first instance
+			await instance1.set('key', 'value');
+
+			await sleep(100);
+
+			// Get the value in the second instance
+			const result = await instance2.get('key');
+			expect(result).toEqual('value');
+
+			// Wait for the value to expire
+			await sleep(700);
+
+			// Get the value in the second instance (it should be expired)
+			const result2 = await instance2.get('key');
+			expect(result2, 'result should have expired').toBeUndefined();
+		});
+
+		test.todo('what\'s the expected behavior when we set an item in the first instance with no ttl, and then get it from the second instance with a ttl?');
+	});
 });
 
 describe('cacheable has method', async () => {
--- a/packages/cacheable/test/index.test.ts
+++ b/packages/cacheable/test/index.test.ts
@@ -397,6 +397,39 @@ describe('cacheable set and get with ttl', async () => {
 			expect(result2, 'result should have expired').toBeUndefined();
 		});
 
+		test('should not return a value from the secondary store or set it in the primary store when the value is expired in the secondary store', async () => {
+			const instance1Primary = new Keyv();
+			const instance2Primary = new Keyv();
+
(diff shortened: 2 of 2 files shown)


TylerEther:
Hi @jaredwray, can you review the test cases for the expected TTL propagation behavior and suggest any missing ones?

There's one test case that I'm not 100% sure of and haven't added yet. In this case, the first instance doesn't have a default TTL, and the client doesn't set one when setting the value. What should happen with the second instance upon retrieval? What happens when the second instance uses a default TTL? I think the second instance's default TTL should be ignored, and the TTL used when the client initially sets the value should always be respected.

jaredwray:
@TylerEther - I havent seen the case where in a project we are running a shared backend and two different ttl's what would be the use case for this? Why wouldn't there be `namespace` on this?

I updated what I believe the use cases are. 

One major note is that they way this works is that Keyv returns an expires and not a ttl. We adhere to the `expires` and set that in memory to build consistency from the initial value.

TylerEther:
@jaredwray I have a setup where multiple programs utilize the same Redis backend for caching. I have one program that writes to the cache and one or more programs that read from the cache.

The problem I'm experiencing is that when the reading program retrieves the value from the secondary store, Redis, it stores the value indefinitely in the primary store, memory.

The TTL setup I currently have is:
- All cacheable and keyv instances across all programs have a default TTL of 10s
- Upon setting the value in the writing program, a specific TTL of 1m is used

koestersc:
hi, chiming in here, as we've just discovered the probably same issue. Also using in-memory as primary store, and redis as shared, secondary store.

I think the issue is this line:
https://github.com/jaredwray/cacheable/blob/a95f7f7f8c2dafcd78fee6a310874a6b03ed974b/packages/cacheable/src/index.ts#L305

the value received from redis, in case the primary store is empty (e.g. when a new instance spins up), returns the unix time in seconds, e.g. `1742573113703`. however, the ttl parameter expects a time-to-live, aka how many seconds/milliseconds the key should be valid. By passing in this unix timestamp, you effectively double the time since 1970, ending up some time in june 2080 for the validity of the in-memory stored key.

TylerEther:
Great catch and diagnosis @koestersc, thank you.

I've committed a fix to the code you outlined and now all test cases are passing.

TylerEther:
Now, we need some test cases for the expected behavior when setting items that don't expire. Since we propagate expiry, I imagine it would be best to also propagate lack-of-expiry.

jaredwray:
@TylerEther - that makes sense. Running the tests

TylerEther:
@jaredwray, I've committed the final tests, and this PR is now ready for review. Thanks!

krilbert:
Amazing today I was implementing this, I came up with
```typescript
secondary.hooks.addHandler(KeyvHooks.POST_GET, ({ key, value }) => {
  const { value: data, expires } = value
  if (!data || !expires) return

  const now = Date.now()
  const ttl = expires - now
  if (ttl > 0) primary.set(key, data, ttl)
})
```
But I guess, I'm not gonna need it anymore 😅
I'm also using in memory as first and redis as a secondary, and as @koestersc mentioned I noticed it on other instances

koestersc:
In case it's relevant for someone:

Until this is resolved, we're manually our own implementation of getting (and setting back) a key.

```
    const memoryCachedToken = await this.cacheClient.primary.get<string>(key, { raw: true });
    if (memoryCachedToken && memoryCachedToken.value) {
      return memoryCachedToken.value;
    }

    const redisCachedToken = await this.cacheClient.secondary?.get<string>(key, { raw: true });
    if (redisCachedToken && redisCachedToken.value) {
      await this.cacheClient.primary.set(
        key,
        redisCachedToken.value,
        redisCachedToken.expires ? redisCachedToken.expires - Date.now() : undefined,
      );
      return redisCachedToken.value;
    }
```

jaredwray:
@TRiLeZ @koestersc @krilbert - this should be released in the next week or so. Thanks so much for the help 🎉

krilbert:
thank you, looking forward 😄

krilbert:
Hello @jaredwray, just a quick question, I didn't know where to put it.
Have you ever consider to expose the raw get? Or has anyone asked for it?

Just wondering because I wanted to do some stale while revalidate (I suppose this could be a common use case) and came up with this (only extend it):

```typescript
export interface CustomCache extends Cacheable {
  get<T>(key: string): Promise<T | undefined>;
  get<T>(key: string, opts?: { raw?: false }): Promise<T | undefined>;
  get<T>(key: string, opts?: { raw: true }): Promise<StoredDataRaw<T> | undefined>;
}

const cache: CustomCache = new Cacheable({ primary, secondary });

cache.get = async <T>(key: string, opts: { raw?: boolean } = {}) => {
  let result: StoredDataRaw<T> | undefined;
  const { raw = false } = opts;

  try {
    await cache.hook(CacheableHooks.BEFORE_GET, key);
    result = await primary?.get(key, { raw: true });
    if (!result && secondary) {
      const rawSecondaryResult = await secondary.get(key, { raw: true });
      if (rawSecondaryResult?.value) {
        result = rawSecondaryResult;
        const ttl = rawSecondaryResult.expires ? rawSecondaryResult.expires - Date.now() : undefined;
        await primary?.set(key, rawSecondaryResult.value, ttl);
      }
    }
    await cache.hook(CacheableHooks.AFTER_GET, { key, result: result?.value });
  } catch (error: unknown) {
    cache.emit(CacheableEvents.ERROR, error);
  }
  return raw ? result : result?.value;
};
```

And with that I mean anyone can do their own implementation, today I came up with something like this:
```ts
interface CacheOptions {
  staleTime?: number | string;
  ttl?: number | string;
}

async function fetchWithCache<T = any>(key: string, fetchFn: () => Promise<T>, options: CacheOptions = {}): Promise<T> {
  // Get raw data from cache
  const cached = await this.cache?.get<T>(key, { raw: true });

  // Fn used to fetch data and refresh the cache
  const refreshCache = async () => {
    try {
      const freshData = await fetchFn();
      await this.cache?.set(key, freshData, options.ttl);
      return freshData;
    } catch (e) {
      if (cached && cached.value) return cached.value;
      throw e;
    }
  };

  if (cached && cached.value) {
    const { value, expires } = cached;
    const remainingTtl = expires ? expires - Date.now() : undefined;
    const staleTime = shorthandToMilliseconds(options.staleTime);

    const isStale = staleTime && remainingTtl && remainingTtl < staleTime;
    if (isStale) {
      // Return stale data immediately and refresh the cache in the background
      refreshCache();
      return value;
    }
    return value;
  }

  // If no cache exists, fetch and cache the result
  return refreshCache();
}
```

Just wonder your thoughts first instead of going directly to create a PR

jaredwray:
> Hello @jaredwray, just a quick question, I didn't know where to put it. Have you ever consider to expose the raw get? Or has anyone asked for it?
> 
> Just wondering because I wanted to do some stale while revalidate (I suppose this could be a common use case) and came up with this (only extend it):
> 
> ```ts
> export interface CustomCache extends Cacheable {
>   get<T>(key: string): Promise<T | undefined>;
>   get<T>(key: string, opts?: { raw?: false }): Promise<T | undefined>;
>   get<T>(key: string, opts?: { raw: true }): Promise<StoredDataRaw<T> | undefined>;
> }
> 
> const cache: CustomCache = new Cacheable({ primary, secondary });
> 
> cache.get = async <T>(key: string, opts: { raw?: boolean } = {}) => {
>   let result: StoredDataRaw<T> | undefined;
>   const { raw = false } = opts;
> 
>   try {
>     await cache.hook(CacheableHooks.BEFORE_GET, key);
>     result = await primary?.get(key, { raw: true });
>     if (!result && secondary) {
>       const rawSecondaryResult = await secondary.get(key, { raw: true });
>       if (rawSecondaryResult?.value) {
>         result = rawSecondaryResult;
>         const ttl = rawSecondaryResult.expires ? rawSecondaryResult.expires - Date.now() : undefined;
>         await primary?.set(key, rawSecondaryResult.value, ttl);
>       }
>     }
>     await cache.hook(CacheableHooks.AFTER_GET, { key, result: result?.value });
>   } catch (error: unknown) {
>     cache.emit(CacheableEvents.ERROR, error);
>   }
>   return raw ? result : result?.value;
> };
> ```
> 
> And with that I mean anyone can do their own implementation, today I came up with something like this:
> 
> ```ts
> interface CacheOptions {
>   staleTime?: number | string;
>   ttl?: number | string;
> }
> 
> async function fetchWithCache<T = any>(key: string, fetchFn: () => Promise<T>, options: CacheOptions = {}): Promise<T> {
>   // Get raw data from cache
>   const cached = await this.cache?.get<T>(key, { raw: true });
> 
>   // Fn used to fetch data and refresh the cache
>   const refreshCache = async () => {
>     try {
>       const freshData = await fetchFn();
>       await this.cache?.set(key, freshData, options.ttl);
>       return freshData;
>     } catch (e) {
>       if (cached && cached.value) return cached.value;
>       throw e;
>     }
>   };
> 
>   if (cached && cached.value) {
>     const { value, expires } = cached;
>     const remainingTtl = expires ? expires - Date.now() : undefined;
>     const staleTime = shorthandToMilliseconds(options.staleTime);
> 
>     const isStale = staleTime && remainingTtl && remainingTtl < staleTime;
>     if (isStale) {
>       // Return stale data immediately and refresh the cache in the background
>       refreshCache();
>       return value;
>     }
>     return value;
>   }
> 
>   // If no cache exists, fetch and cache the result
>   return refreshCache();
> }
> ```
> 
> Just wonder your thoughts first instead of going directly to create a PR

We have been thinking about adding in `getRaw()` as a function or within the options like you show here. Can you create this a new issue so we can track it as an enhancement we need to add?

jaredwray:
@TRiLeZ, @krilbert, @koestersc - We are making the expires work better with an updated fix. Just wanted to send an FYI on this and will be adding in documentation on it. Please follow along here: https://github.com/jaredwray/cacheable/issues/1077

krilbert:
Uff awesome, first of all sorry for the late response, I got distracted with work and PTOs 😳 
Happy to add the issue :D great job everyone :D

PR #1080, opened 2025-04-10
cacheable - fix: deleting a key should delete entry from hash cache

JohnICB:
**Please check if the PR fulfils these requirements**
- [ ] Tests for the changes have been added (for bug fixes/features) with 100% code coverage.

This PR introduces a bug fix and a small improvement.

### Bug Fix
The bug fix is related to the _hashCache variable not having its key deleted when it's deleted from the store. This would result in increasing memory consumption over time.
A use case: you receive a lot of IDs (1000s / second) and you want to cache them for a short period of time (using TTL). They are deleted from the store, but not from the hashCache, thus increasing the memory consumption indefinitely.

I could also add this change under an opt-in flag if the current behaviour is intended (e.g. if a key is deleted, but later you check it again and you don't want to compute the hash again). This would be fine for a reduced number of unique keys (< 10k? )

### Improvement
The hashKey function would compute the hash over and over for values stored in storage index 0.

### DIFF
--- a/packages/cacheable/src/memory.ts
+++ b/packages/cacheable/src/memory.ts
@@ -365,6 +365,7 @@ export class CacheableMemory extends Hookified {
 	public delete(key: string): void {
 		const store = this.getStore(key);
 		store.delete(key);
+		this._hashCache.delete(key);
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


jaredwray:
@JohnICB thanks so much for finding this bug and fixing it! This should get released in the next couple weeks.

Issue #1079, opened 2025-04-09
[cache-manger]: Object of type Uint8Array is not returned  correctly from cache


re-kfa:
**Describe the bug**
When saving an object of type Uint8Array in a cache it will not be returned as an Uint8Array-object but as a plain object.
This is a regression introduced in V6. I updated today from V5 and noticed this change in behaviour.

**How To Reproduce**
``` 
import { createCache } from "cache-manager";

const memoryCache = createCache();

const getBinaryObjectCached = (key) => {
	return memoryCache.wrap(key, () => new Uint8Array(["1", "2", "3"]), 300);
};

const key = "key";

let content = await getBinaryObjectCached(key);
console.log(content);

let content2 = await getBinaryObjectCached(key);
console.log(content2);
```

***Expected Result***
```
Uint8Array(3) [ 1, 2, 3 ]
Uint8Array(3) [ 1, 2, 3 ]
```

***Actual Result***
```
Uint8Array(3) [ 1, 2, 3 ]
{ '0': 1, '1': 2, '2': 3 }
```

jaredwray:
@re-kfa - it looks like you are using the default settings. This could be based on using Keyv as our primary storage adapter layer even for in-memory. Keyv does do serialization by default even for in-memory based on legacy settings. Have you tried removing the serialization?

```js
import { createCache } from 'cache-manager';
import { Keyv } from 'keyv';

const keyv = new Keyv();
keyv.serialization = undefined;
keyv.deserialization = undefined;

const cache = createCache({
  stores: [keyv],
});
```

re-kfa:
@jaredwray - thanks for your quick answer. removing the serialization indeed solves the problem.
This is the working code:
```
import { createCache } from "cache-manager";
import { Keyv } from "keyv";

const keyv = new Keyv();
keyv.serialize = undefined;
keyv.deserialize = undefined;

const memoryCache = createCache({
	stores: [keyv],
});
```

Would it make sense to add something about this behaviour change to the `Migration from v5 to v6` section in the Readme?

jaredwray:
> [@jaredwray](https://github.com/jaredwray) - thanks for your quick answer. removing the serialization indeed solves the problem. This is the working code:
> 
> ```
> import { createCache } from "cache-manager";
> import { Keyv } from "keyv";
> 
> const keyv = new Keyv();
> keyv.serialize = undefined;
> keyv.deserialize = undefined;
> 
> const memoryCache = createCache({
[...]

Issue #843, opened 2024-10-18
Date objects are stored as strings


lugrinder:
**Describe the bug**
Date properties are stored as string instead of object. 
On many ORM, date/time fields are returned as Date objects, so cache its results causes this issue.
I think that versions before `keyv` works, but not really sure, I've detected the issue after updating to v6.
If this will be the default behavior, please fix the documentation and add typescript type guard to ensure that only valid values can be cached, or better, transform the result type for `wrap` for all Date properties to be the original returned value (Date) and string: Date | string.

**How To Reproduce (best to provide workable code or tests!)**
Cache any value with a Date property.

lugrinder:
I can confirm that v5 stores Date as object and don't have this issue. I've fixed it, adding the next typescript type to the returned value of `wrap` that have `Date` properties, so typescript warns about this if I've tried to use as `Date`:

```typescript
type StringifyObject<T> = {
    [P in keyof T]: Extract<T[P], object> extends never ? T[P] : T[P] | string
}
```

At least, add something similar to the returned `wrap` value.

jaredwray:
@lugrinder - thanks for letting us know. Can you tell me what package you are seeing this on? Is this `cache-manager`?

Also, can you give me an example of where you are using this code to fix it?

lugrinder:
Yes, `cache-manager`.
You can see here a [running example](https://codesandbox.io/p/devbox/amazing-wildflower-zyjl6j).

jaredwray:
got it. I see. Where are you adding this code to fix it?

```
type StringifyObject<T> = {
    [P in keyof T]: Extract<T[P], object> extends never ? T[P] : T[P] | string
}
```

lugrinder:
I'm not modified the original code, I've done something like this:

```typescript
async function getData(name: string) {
    const result = await cache.wrap(name, async () => {
        return new Promise((resolve) => {
            resolve({
                a: 'hello',
                b: new Date
            })
        })
    })
    return result as StringifyObject<typeof result>
}
```
If you do it inside `wrap`, there is no need to add in all calls to `wrap`, you can infer from the wrapped promise.
The question is, if not possible to store objects as v5, at least make a union for the transformed types and ensure that typescript will warn if you don't take care.

jaredwray:
@lugrinder - thanks for the info and I am going to see what we can do to help with this.

jaredwray:
[...]

PR #1086, opened 2025-04-14
cache-manager - fix: in-memory defaults on createCache with no serial…

jaredwray:

With the migration to `Keyv` as the core storage adapter in `cache-manager v6` it has caused multiple issues with the simple in memory because Keyv by default does *serialization*. Here is an example:

```js
import { createCache } from 'cache-manager';
const cache = createCache();
```

This creates a simple in memory cache and with `v5` it had no serialization. To resolve this we need to set `keyv.serialize` and `keyv.deserialize` to `undefined`. 

This fix makes it so the default Keyv if you specify no store will handle this by default. 

#843, #1079

### DIFF
--- a/packages/cache-manager/README.md
+++ b/packages/cache-manager/README.md
@@ -38,14 +38,26 @@ const memoryCache = createCache(memoryStore({
 }));
 ```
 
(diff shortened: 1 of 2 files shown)


lcswillems:
Hey @jaredwray , looks great! Do you have an idea when it will be released? :pray:

PR #1221, opened 2025-07-22
cache-manager - feat: `mget` and `mset` call `getMany` and `setMany`

### LINKED ISSUE #1040: MGET/MSET implementation in cache-manager not using MGET/MSET implementation in keyv stores
**Is your feature request related to a problem? Please describe.**
Current implementation of `mget` and `mset` in cache-manager end up using multiple calls to `get` and `set` methods in underlying Keyv stores, wrapped in a `Promise.all`. In clients like Redis, which supports native MGET and MSET commands, this is unefficient. 

Example of current implementations:

```
  const mget = async (keys) => {
    const result = [];
    for (const key of keys) {
      const data = await get(key);
      result.push(data);
    }
    return result;
  };

```

```
  const mset = async (stores2, list) => {
    const items = list.map(({ key, value, ttl: ttl2 }) => ({ key, value, ttl: ttl2 }));
    try {
      const promises = [];
      for (const item of list) {
        promises.push(stores2.map(async (store) => store.set(item.key, item.value, item.ttl)));
      }
      if (nonBlocking) {
        Promise.all(promises);
        eventEmitter.emit("mset", { list });
        return list;
      }
      await Promise.all(promises);
      eventEmitter.emit("mset", { list });
      return list;
    } catch (error) {
      eventEmitter.emit("mset", { list, error });
      return Promise.reject(error);
    }
  };

```

**Describe the solution you'd like**
The underlying Keyv stores already have getMany and setMany methods, which could be used in the mget and mset method in Cache Manager, instead of calling N times the single implementations.

nathan-knight:

Feature/bugfix
- `mget` calls `getMany`
- `mset` calls `setMany`
- `mset` uses the default ttl for items if the ttl isn't specified (this was a bug)

This PR resolves #1040 (I couldn't find any changes that would have resolved it despite it being closed for a few months)

### DIFF
--- a/packages/cacheable/package.json
+++ b/packages/cacheable/package.json
@@ -28,22 +28,22 @@
 		"clean": "rimraf ./dist ./coverage ./node_modules"
 	},
 	"devDependencies": {
-		"@faker-js/faker": "^9.8.0",
-		"@keyv/redis": "^4.5.0",
-		"@keyv/valkey": "^1.0.6",
+		"@faker-js/faker": "^9.9.0",
+		"@keyv/redis": "^5.0.0",
+		"@keyv/valkey": "^1.0.7",
 		"@types/eslint": "^9.6.1",
-		"@types/node": "^24.0.7",
+		"@types/node": "^24.1.0",
 		"@vitest/coverage-v8": "^3.2.4",
 		"lru-cache": "^11.1.0",
 		"rimraf": "^6.0.1",
 		"tsup": "^8.5.0",
 		"typescript": "^5.8.3",
 		"vitest": "^3.2.4",
-		"xo": "^1.1.1"
+		"xo": "^1.2.1"
 	},
 	"dependencies": {
 		"hookified": "^1.10.0",
-		"keyv": "^5.3.4"
+		"keyv": "^5.4.0"
 	},
 	"keywords": [
 		"cacheable",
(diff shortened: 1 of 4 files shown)


packages/cache-manager/src/index.ts
```diff
@@ -122,11 +124,28 @@ export const createCache = (options?: CreateCacheOptions): Cache => {
...
+			try {
+				result = await Promise.race(stores.map(async store => store.getMany(keys)));
+				eventEmitter.emit('mget', {keys, values: result});
+			} catch (error) {
+				eventEmitter.emit('mget', {keys, error});
+			}
+		} else {
+			for (const store of stores) {
+				try {
+					const cacheValue = await store.getMany<T>(keys);
```
nathan-knight, line 139: I wasn't sure if it should take the first successful response entirely or if it should go to each store and merge the results if an earlier store didn't return something

jaredwray: for nonBlocking you should take the first store response. I believe the pattern that we have been going with for normal mode is we loop through all of them and merge results. Correct?

jaredwray:
@nathan-knight - looks like you have a lint error

```
  src/index.ts:209:38
    ✖  209:38  Functions that return promises must be async.  @typescript-eslint/promise-function-async
```

packages/cache-manager/src/index.ts
nathan-knight: Pushed that change and fixed the lint error

jaredwray:
@nathan-knight - thanks so much and I will get this published today or tomorrow.

Issue #1617, opened 2026-04-15
node-cache: Keys are not removed after expiration until get


KristjanTammekivi:
**Describe the bug**
.has returns true even if a key is expired. This differs from how node-cache works and breaks the following pattern

```typescript
const getCached = (query, ttl) => {
  if (cache.has(query.name)) {
    return cache.get(query.name); // returns undefined if value has expired
  }
  // execute query and set cache
}
```

Since in README.md it says that @cacheable/node-cache is compatible with node-cache I'm filing this as a bug.

**How To Reproduce (best to provide workable code or tests!)**
https://github.com/KristjanTammekivi/cacheable-bug-report/blob/main/index.ts

```[2026-04-15T13:17:28.923Z] NodeCache value: Hello, World!
[2026-04-15T13:17:28.923Z] Cacheable value: Hello, Cacheable!
[2026-04-15T13:17:29.024Z] NodeCache has key: false
[2026-04-15T13:17:29.024Z] Cacheable has key: true
[2026-04-15T13:17:29.024Z] NodeCache value after expiration: undefined
[2026-04-15T13:17:29.024Z] Cacheable value after expiration: undefined
[2026-04-15T13:17:29.025Z] NodeCache has key after get: false
[2026-04-15T13:17:29.025Z] Cacheable has key after get: false
```

jaredwray:
@KristjanTammekivi this is fixed and will be released in a week or so

PR #1618, opened 2026-04-15
node-cache - fix: has was not removing expired keys

jaredwray:
has() previously returned true for expired keys, breaking the
has()/get() pattern used with the original node-cache. It now checks
the TTL, emits "expired", and removes the key (when deleteOnExpire is
enabled), matching node-cache's behavior.


jaredwray:
https://github.com/jaredwray/cacheable/issues/1617

PR #1632, opened 2026-05-07
@cacheable/benchmark - fix: upgrade @faker-js/faker to 10.4.0 (breaking)

jaredwray:
## Summary
- Upgrade `@faker-js/faker` from 9.8.0 to 10.4.0 in `@cacheable/benchmark` to align with the workspace root.

## Notes
- Faker 10 dropped CommonJS support and requires Node 20+; the benchmark package is private and only uses the modern ESM API, so call sites are unaffected.

## Test plan
- [x] `pnpm install` succeeds
- [x] `pnpm --filter @cacheable/benchmark test` passes (lint only — package has no tests)

### DIFF
--- a/packages/benchmark/package.json
+++ b/packages/benchmark/package.json
@@ -48,7 +48,7 @@
 	},
 	"dependencies": {
 		"@cacheable/memory": "workspace:^",
-		"@faker-js/faker": "^9.8.0",
+		"@faker-js/faker": "^10.4.0",
 		"@monstermann/tinybench-pretty-printer": "^0.1.0",
 		"bentocache": "^1.4.0",
 		"cache-manager": "workspace:^",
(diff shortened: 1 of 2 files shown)


jaredwray:
Re: gemini-code-assist's review — `faker.helpers.rangeToNumber` was **not** removed in faker v10. It's still in the public API of `@faker-js/faker@10.4.0` (added in 8.0.0, still present in v10 as documented in `dist/airline-eVQV6kbz.d.ts`).

Verified at runtime against the installed 10.4.0 build:
```
faker version: 10.4.0
rangeToNumber type: function
result: 5
```

Also smoke-tested `ObjectGenerator` end-to-end:
```
user: true
product: true
order: true productIds count: 3
review: 4
```

No code changes needed.

PR #1645, opened 2026-05-26
feat(cacheable, memory): add maxTtl option to cap maximum time-to-live

jaredwray:
## Summary

Adds a `maxTtl` option to both `Cacheable` and `CacheableMemory` that enforces an upper bound on any TTL in the cache. When set, per-entry TTLs exceeding `maxTtl` are capped, and entries with no TTL (that would otherwise live indefinitely) are also bounded. This is useful for guaranteeing that no cache entry outlives a maximum duration regardless of what TTL is passed to individual `set()` calls.

## Changes

- Add `maxTtl` option to `CacheableMemoryOptions` type and `CacheableMemory` class with getter/setter, constructor handling, and enforcement in `set()`
- Add `maxTtl` option to `CacheableOptions` type and `Cacheable` class with getter/setter, constructor handling, and enforcement in `set()` and `setMany()` (both primary and secondary stores)
- Add comprehensive test suites for both packages (100% coverage on memory package)
- Update README documentation for both `@cacheable/memory` and `cacheable` packages with usage examples and API reference

## Verification

- [x] `pnpm build` passes
- [x] All memory package tests pass (113/113, 100% coverage)
- [x] All non-Redis cacheable tests pass (120/120)
- [x] Lint passes (`biome check --error-on-warnings`)

## Test plan

```bash
cd packages/memory && pnpm test   # 113 tests, 100% coverage
cd packages/cacheable && npx vitest run test/index.test.ts  # 120 pass (12 Redis-dependent skip without Redis)
```

Key scenarios tested:
- `maxTtl` defaults to `undefined`
- Setting via constructor (number and string shorthand)
- Setting via property setter
- Disabling by setting to `undefined`, `0`, or negative
- Capping TTL when it exceeds `maxTtl`
- Not capping when TTL is within `maxTtl`
- Enforcing on entries with no TTL
- Enforcing when default TTL exceeds `maxTtl`
- Working with shorthand strings (e.g., `'1s'`, `'1h'`)
- Actual expiration at `maxTtl` boundary
- Enforcement on `setMany()`
- Enforcement on secondary store
- Enforcement with `SetOptions` object (both `ttl` and `expire` fields)

https://claude.ai/code/session_01Gdiqk2rKNc1Vji61zm3qmy

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gdiqk2rKNc1Vji61zm3qmy)_

### DIFF
--- a/packages/cacheable/README.md
+++ b/packages/cacheable/README.md
@@ -32,6 +32,7 @@
 * [Storage Tiering and Caching](#storage-tiering-and-caching)
 * [TTL Propagation and Storage Tiering](#ttl-propagation-and-storage-tiering)
 * [Shorthand for Time to Live (ttl)](#shorthand-for-time-to-live-ttl)
+* [Maximum Time to Live (maxTtl)](#maximum-time-to-live-maxttl)
 * [Iteration on Primary and Secondary Stores](#iteration-on-primary-and-secondary-stores)
 * [Non-Blocking Operations](#non-blocking-operations)
 * [Non-Blocking with @keyv/redis](#non-blocking-with-keyvredis)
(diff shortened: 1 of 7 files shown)


jaredwray:
## Code Review Response

An internal staff-engineer-grade code review was run against this branch. Here's each finding and the action taken:

---

### ⚠️ Major Findings

**1. BEFORE_SET hook can bypass maxTtl on primary store** (`packages/cacheable/src/index.ts`)

The `BEFORE_SET` hook could mutate `item.ttl` after the maxTtl cap was applied, and the uncapped value would be written to the primary store. The secondary store was correctly re-capped, making behavior inconsistent.

**✅ Fixed in `3e14a6f`.** Added `item.ttl = this.capTtl(item.ttl, maxTtlMs)` after the hook call, before writing to the primary store. Added a test (`should re-cap ttl after BEFORE_SET hook overrides it`) confirming maxTtl holds even when hooks override the TTL.

---

**2. Sync subscriber bypasses maxTtl** (`packages/cacheable/src/sync.ts`)

Sync writes directly to Keyv via `storage.set()` without passing through `Cacheable.set()`, so a remote instance's TTL isn't re-capped locally.

**⏭️ No change.** The originating instance already caps the TTL via `maxTtl` before publishing the sync event, so the published TTL is already capped. If instances have different `maxTtl` configs, that's an intentional operational choice. Changing sync to route through `Cacheable.set()` would be architecturally significant (re-triggering hooks, risk of sync loops) and warrants a separate discussion if needed.

---

**3. Commit hygiene — branch mixes unrelated changes**

**⏭️ False positive.** The NodeCacheStore overhaul, tsdown migration, and CI changes are pre-existing commits on `main` (`bf3ea48`, `1508695`) that appear in the three-dot diff. This branch contains only 2 commits: the `maxTtl` feature and the hook-bypass fix.

---

**4. Extra `get()` on every `set()` for stats in NodeCacheStore**

**⏭️ Out of scope.** This is pre-existing code from PR #1643 (NodeCacheStore enhancement), not part of the maxTtl feature.

---

### 💡 Minor Findings

**5–6. `shorthandToMilliseconds` / `shorthandToTime` called on every `set()`**

Regex parsing happens on each call when maxTtl is a string like `'1h'`.

**⏭️ No change.** The regex parse is a single match against a short pattern — negligible compared to the async store operations that follow. Pre-resolving would add complexity (shadow `_maxTtlMs` field kept in sync) for a micro-optimization unlikely to show in benchmarks. This is consistent with how the existing `ttl` property works throughout the codebase.

---

**7. `mset` fans out 2N network calls in NodeCacheStore**

**⏭️ Out of scope.** Pre-existing code from a different PR, not part of the maxTtl feature.

---

### Coverage

Codecov confirms **100% project coverage maintained** — all modified and coverable lines are covered by tests (+43 lines, +15 branches).

---

PR #1650, opened 2026-06-02
fix(node-cache): Fixing issues with metrics reporting

jmjones88:
- [x] Followed the [Contributing](../blob/main/CONTRIBUTING.md) guidelines and [Code of Conduct](../blob/main/CODE_OF_CONDUCT.md)

This fixes multiple metrics-accounting issues that caused inaccurate ksize and vsize reporting over time.

The changes ensure that:
- Overwriting existing keys does not inflate size metrics.
- take() correctly removes keys even when cached values are falsy.
- NodeCacheStore correctly decrements value-size metrics for undefined values on overwrite, delete, and expiry.


jaredwray:
@jmjones88 - thanks for the fix. We should have this live soon

PR #1655, opened 2026-06-12
feat(cacheable): tag-based invalidation via CacheTags

### LINKED ISSUE #1640: Tag based invalidation
Picking up from #120. That was on the old cache-manager and closed without anything landing. The rewrite has hooks and L1/L2 now, which I think gets around Bryan's "not generic enough" concern.

Use case is content caches: one upstream entity referenced by many cache entries, need to purge all of them when it changes. Right now that means walking the dependency graph or maintaining a Redis SET by hand. Same thing Symfony's TagAwareAdapter and Next.js `revalidateTag` do.

```ts
await cache.set(key, value, {tags: ['entity:42']})
await cache.invalidateTag('entity:42')
```

The `BEFORE_SET` / `AFTER_SET` hooks look like the right spot. Tag index can just be cache entries under a reserved prefix, no adapter changes. Would need to clear both primary and secondary, and probably go through `sync` for multi-instance.

Happy to PR if you're up for it.

jaredwray:
Implements tag-based invalidation for `Cacheable`, closing out the integration step from #1640 on top of the `CacheTags` service merged in #1646.

## API

All tag functionality lives on the `tags` service (a `CacheTags` instance from `@cacheable/utils`), created by default in the constructor. Tags must be explicitly enabled with `tags: true` (or `cache.tags.enabled = true`):

```javascript
const cache = new Cacheable({ tags: true });

await cache.set('page:/products', html, { ttl: '10m', tags: ['entity:42', 'collection:products'] });
await cache.setMany([{ key: 'user:1', value, tags: ['users'] }]);

await cache.tags.invalidateTag('entity:42');
await cache.tags.invalidateTags(['users', 'org:7']);

await cache.get('page:/products'); // undefined

await cache.tags.getTags('user:1'); // ['users']
cache.tags.enabled; // whether the tag service is on
```

`set` keeps full back-compat with `ttl` as the third argument; the options form also honors a per-call `nonBlocking` override (the existing-but-unused `SetOptions` type).

## How it works

- Lazy, constant-time invalidation as discussed in #1640: `invalidateTag` bumps the tag's version counter; no scan-and-delete.
- On `get` / `getMany`, tagged entries whose snapshot no longer matches the live versions are treated as misses and **removed from both primary and secondary stores**, with a `delete` published via sync when enabled — so other instances purge their L1 copies too. `getMany` uses `CacheTags.getStaleKeys`, which checks any number of keys with two batched store reads.
- Tag metadata lives in the **secondary store when configured** (shared invalidation across instances), otherwise the primary store. Snapshots are written with the same TTL as the value copy in that store so they expire together.
- The `CacheTags` service is **self-contained**: it owns its `enabled` state and handles non-blocking fire-and-forget writes internally, reporting failures via an `onError` callback that `Cacheable` wires to its `error` event.
- The service is **disabled by default and never enables itself** — untagged workloads pay zero extra reads, and behavior stays consistent across distributed instances. While disabled, all tag operations are no-ops; enable it with `tags: true` on every instance that shares the store.
- Setting or deleting a key without tags clears any previous snapshot so a stale snapshot can never invalidate a fresh value.

## `@cacheable/utils` additions (`CacheTags`)

- `enabled` property — all methods no-op while disabled; must be explicitly enabled.
- `isKeyStale(key)` — `true` only when a snapshot exists and a tag version changed, so it is safe to call for every lookup including never-tagged keys (unlike `isKeyFresh`, which reports unknown keys as not fresh).
- `getStaleKeys(keys)` — batch staleness check using two store reads total.
- `getTags(key)` — returns a key's tags or `undefined`.
- `removeKeys(keys)` — batched snapshot removal via `deleteMany`.
- `nonBlocking` option on `setKeyTags` / `removeKey` / `removeKeys` with an `onError` constructor callback for fire-and-forget failures.

## Tests & docs

- 29 tests in `packages/cacheable/test/tags.test.ts` (single store, shared secondary across instances, non-blocking mode, snapshot cleanup, TTL expiry, error events) and 24 `CacheTags` tests in `packages/utils/test/cache-tags.test.ts`. All new code paths are covered.
- README sections added/updated for both packages (`Tag Based Invalidation` in cacheable; `enabled`, `isKeyStale`, `getStaleKeys`, `getTags`, `removeKeys`, and non-blocking docs in utils), plus options/API listings.

Note: tag version counters are intentionally stored without TTL (they are the source of truth); the lifecycle/cleanup concern raised late in #1640 is orthogonal to this integration and can be tracked separately.

Closes #1640

https://claude.ai/code/session_01CEUJ2ZvfykEnuUfL43PobY

### DIFF
--- a/packages/cacheable/README.md
+++ b/packages/cacheable/README.md
@@ -646,7 +646,7 @@ You can associate cache entries with tags and later invalidate every entry that 
 ```javascript
 import { Cacheable } from 'cacheable';
 
-const cache = new Cacheable();
+const cache = new Cacheable({ tags: true });
 
 await cache.set('page:/products', html, { ttl: '10m', tags: ['entity:42', 'collection:products'] });
 await cache.set('page:/products/42', detailHtml, { ttl: '10m', tags: ['entity:42'] });
@@ -677,8 +677,8 @@ Tag metadata is stored in the secondary store when one is configured, otherwise 
 import { Cacheable } from 'cacheable';
 import KeyvRedis from '@keyv/redis';
 
-const writer = new Cacheable({ secondary: new KeyvRedis('redis://localhost:6379') });
-// instances that only read tagged entries should opt in with `tags: true`
+// enable tags on every instance that shares the store - writers and readers
+const writer = new Cacheable({ secondary: new KeyvRedis('redis://localhost:6379'), tags: true });
 const reader = new Cacheable({ secondary: new KeyvRedis('redis://localhost:6379'), tags: true });
 
 await writer.set('page:/products', html, { tags: ['entity:42'] });
@@ -686,7 +686,7 @@ await writer.tags.invalidateTag('entity:42');
 await reader.get('page:/products'); // undefined - stale copy is also purged from reader's primary
 ```
 
-The tag service starts disabled so untagged workloads pay no extra cost. It enables automatically the first time a tag is written (setting a value with `tags` or calling `tags.invalidateTag` / `tags.invalidateTags`), or explicitly via the `tags: true` option or the `tags.enabled` property. When using tags across multiple instances that share a store, set `tags: true` on all of them so every instance honors invalidations and cleans up tag snapshots consistently.
+The tag service is disabled by default so untagged workloads pay no extra cost, and you have to turn it on to use it — either with the `tags: true` option or by setting `cache.tags.enabled = true`. While disabled, all tag operations are no-ops: values set with `tags` are stored without tag tracking and invalidations have no effect. The service never enables itself, which keeps behavior predictable across distributed instances — enable it on every instance that shares the store (writers and readers) so invalidations are honored and tag snapshots are cleaned up consistently.
 
 The full `CacheTags` API is available on the service:
 
@@ -708,7 +708,7 @@ The following options are available for you to configure `cacheable`:
 * `maxTtl`: The maximum time to live for any cache entry. When set, TTLs exceeding this value are capped. Enforced on both primary and secondary stores. Default is `undefined` (no maximum).
 * `namespace`: The namespace for the cache. Default is `undefined`.
 * `cacheId`: A unique identifier for this cache instance. Used for sync filtering. Default is a random string.
-* `tags`: Enables the tag service so tag-based invalidation freshness checks run on `get` / `getMany`. The service also enables automatically the first time a tag is written on the instance. Default is `false`.
+* `tags`: Enables the tag service so tag-based invalidation can be used and freshness checks run on `get` / `getMany`. Tags must be explicitly enabled — while disabled, all tag operations are no-ops. Default is `false`.
 * `sync`: Enable distributed cache synchronization. Can be:
   - `CacheableSync` instance
   - `CacheableSyncOptions` object with `{ qified: MessageProvider | MessageProvider[] | Qified }`
--- a/packages/cacheable/README.md
+++ b/packages/cacheable/README.md
@@ -652,7 +652,7 @@ await cache.set('page:/products', html, { ttl: '10m', tags: ['entity:42', 'colle
 await cache.set('page:/products/42', detailHtml, { ttl: '10m', tags: ['entity:42'] });
 
 // entity 42 changed - purge everything that referenced it
-await cache.invalidateTag('entity:42');
+await cache.tags.invalidateTag('entity:42');
 
 await cache.get('page:/products'); // undefined
 await cache.get('page:/products/42'); // undefined
@@ -666,10 +666,10 @@ await cache.setMany([
   { key: 'user:2', value: userTwo, tags: ['users', 'org:7'] },
 ]);
 
-await cache.invalidateTags(['users', 'org:7']);
+await cache.tags.invalidateTags(['users', 'org:7']);
 ```
 
-Invalidation is powered by the `CacheTags` service from [`@cacheable/utils`](https://npmjs.com/package/@cacheable/utils) and uses a lazy, constant-time model: `invalidateTag` simply bumps a version counter for the tag, no matter how many entries reference it. Each tagged entry stores a snapshot of its tags' versions, and on the next `get` / `getMany` the snapshot is compared to the live versions. If any tag has been bumped since, the entry is treated as a miss and removed from both the primary and secondary stores (and a `delete` is published via [sync](#cacheablesync---distributed-updates) when enabled). The trade-off is one additional tag-store read per cache lookup while tag support is enabled.
+Tag functionality lives on the `tags` service — an instance of the `CacheTags` class from [`@cacheable/utils`](https://npmjs.com/package/@cacheable/utils) that is created by default in the constructor. Invalidation uses a lazy, constant-time model: `invalidateTag` simply bumps a version counter for the tag, no matter how many entries reference it. Each tagged entry stores a snapshot of its tags' versions, and on the next `get` / `getMany` the snapshot is compared to the live versions. If any tag has been bumped since, the entry is treated as a miss and removed from both the primary and secondary stores (and a `delete` is published via [sync](#cacheablesync---distributed-updates) when enabled). The trade-off is one additional tag-store read per cache lookup while the tag service is enabled.
 
 Tag metadata is stored in the secondary store when one is configured, otherwise in the primary store. With a shared secondary store (such as Redis), an invalidation performed by one instance is seen by every instance:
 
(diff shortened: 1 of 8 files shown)


packages/utils/src/cache-tags.ts
```diff
@@ -217,6 +217,49 @@ export class CacheTags {
...
+	 * the key has no snapshot).
+	 */
+	public async isKeyStale(key: string): Promise<boolean> {
+		const entry = await this._store.get<KeyTagEntry>(this.keyEntryKey(key));
+		if (!entry?.tags) {
+			return false;
+		}
+
+		const tags = Object.keys(entry.tags);
+		const currentVersions = await this.getTagVersions(tags);
```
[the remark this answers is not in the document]
jaredwray: Keyv v5's `get` natively supports array keys — it checks `Array.isArray(key)` and delegates to `getMany` internally, so this works with all adapters. `getTagVersions` is also pre-existing code from #1646 rather than part of this change, so leaving it as-is here.

The other three suggestions (batch `removeKeys` on `CacheTags`, batched snapshot removal, and batched stale-key deletion in `getManyRaw`) were good calls and are implemented in 5120c75.

---

packages/cacheable/src/index.ts
```diff
@@ -38,6 +44,8 @@ export class Cacheable extends Hookified {
 	private _namespace?: string | (() => string);
 	private _cacheId: string = Math.random().toString(36).slice(2);
 	private _sync?: CacheableSync;
+	private _cacheTags?: CacheTags;
```
jaredwray: just name this _tags and make a public property called `tags`

packages/cacheable/src/index.ts
```diff
@@ -301,6 +315,36 @@ export class Cacheable extends Hookified {
...
+	 *
+	 * [Learn more about tag-based invalidation](https://cacheable.org/docs/cacheable/#tag-based-invalidation).
+	 *
+	 * @returns {CacheTags} The tag service for the cacheable instance
+	 */
+	public get tags(): CacheTags {
+		this._tagsEnabled = true;
+		this._cacheTags ??= new CacheTags({
+			store: this._secondary ?? this._primary,
+		});
```
jaredwray: this should be done by default in the constructor

packages/cacheable/src/index.ts
```diff
@@ -38,6 +44,8 @@ export class Cacheable extends Hookified {
 	private _namespace?: string | (() => string);
 	private _cacheId: string = Math.random().toString(36).slice(2);
 	private _sync?: CacheableSync;
+	private _cacheTags?: CacheTags;
+	private _tagsEnabled = false;
```
jaredwray: This should be a property on the CacheTags class as we will need to use it everywhere. Update the CacheTags service to understand if it is enabled or not. Then isKeyFresh and calls can be more no-op inside the service itself. Better to be self contained

packages/cacheable/src/index.ts
```diff
@@ -502,6 +555,21 @@ export class Cacheable extends Hookified {
...
+						if (result[i] && (await this.tags.isKeyStale(key))) {
+							staleKeys.push(key);
+							result[i] = undefined;
+						}
+					}),
+				);
+				if (staleKeys.length > 0) {
+					await this.deleteMany(staleKeys);
+				}
+			}
```
jaredwray: this should get moved to the CacheTags class. Look at a better architecture for this

packages/cacheable/src/index.ts
```diff
@@ -827,6 +924,64 @@ export class Cacheable extends Hookified {
...
+	 * ```typescript
+	 * const cache = new Cacheable();
+	 * await cache.set('page:/products', html, { ttl: '10m', tags: ['entity:42'] });
+	 * await cache.invalidateTag('entity:42');
+	 * await cache.get('page:/products'); // undefined
+	 * ```
+	 */
+	public async invalidateTag(tag: string): Promise<void> {
+		await this.invalidateTags([tag]);
+	}
```
jaredwray: we should have this as part of the CacheTags class

packages/cacheable/src/index.ts
```diff
@@ -827,6 +924,64 @@ export class Cacheable extends Hookified {
...
+		if (tags.length === 0) {
+			return;
+		}
+
+		try {
+			await this.tags.invalidateTags(tags);
+		} catch (error: unknown) {
+			this.emit(CacheableEvents.ERROR, error);
+		}
+	}
```
jaredwray: should be part of the cachetags class. from cacheable it should be we access this function like cacheable.tags.invalidateTags()

packages/cacheable/src/index.ts
```diff
@@ -827,6 +924,64 @@ export class Cacheable extends Hookified {
...
+	 * @returns {Promise<string[] | undefined>} The tags for the key
+	 */
+	public async getTags(key: string): Promise<string[] | undefined> {
+		try {
+			return await this.tags.getKeyTags(key);
+		} catch (error: unknown) {
+			this.emit(CacheableEvents.ERROR, error);
+			return undefined;
+		}
+	}
```
jaredwray: should be cacheable.tags.getTags

packages/cacheable/src/index.ts
```diff
@@ -1005,6 +1160,90 @@ export class Cacheable extends Hookified {
...
+		const promise = this.tags.setKeyTags(key, tags, { ttl });
+		if (nonBlocking) {
+			promise.catch((error) => {
+				this.emit(CacheableEvents.ERROR, error);
+			});
+			return;
+		}
+
+		await promise;
+	}
```
jaredwray: can we move this to cachetags

packages/cacheable/src/index.ts
```diff
@@ -1005,6 +1160,90 @@ export class Cacheable extends Hookified {
...
+		const promise = this.tags.removeKeys(keys);
+		if (nonBlocking) {
+			promise.catch((error) => {
+				this.emit(CacheableEvents.ERROR, error);
+			});
+			return;
+		}
+
+		await promise;
+	}
```
jaredwray: move this to cacheTags

packages/cacheable/src/index.ts
jaredwray: All addressed in d855c2c — CacheTags is now self-contained and Cacheable just delegates:

- `enabled` is a property on `CacheTags`: read methods (`isKeyFresh`, `isKeyStale`, `getStaleKeys`, `getTags`, `getKeysByTag`) no-op while disabled, and tag writes (`setKeyTags`, `invalidateTag`, `invalidateTags`) auto-enable the service.
- This block moved into `CacheTags.getStaleKeys(keys)`, which checks all keys with two batched store reads total (one for snapshots, one for the union of tag versions) instead of two per key; `getManyRaw` just calls it and batch-deletes the result.
- Non-blocking handling moved into the service: `setKeyTags`/`removeKey`/`removeKeys` take a `nonBlocking` option and report fire-and-forget failures via an `onError` constructor callback (duck-typed like `Stats`, so utils stays free of an emitter dependency) — Cacheable wires it to its `error` event.
- Field renamed to `_tags` with a public `tags` getter; the service is created by default in the constructor (and recreated when primary/secondary change, preserving enabled state).
- Top-level `invalidateTag`/`invalidateTags`/`getTags`/`tagsEnabled` removed from Cacheable — everything is `cacheable.tags.*` now, with `getKeyTags` renamed to `getTags`.

---

packages/cacheable/src/index.ts
```diff
@@ -502,6 +564,21 @@ export class Cacheable extends Hookified {
...
+				const staleKeys = await this._tags.getStaleKeys(presentKeys);
+				if (staleKeys.length > 0) {
+					const staleSet = new Set(staleKeys);
+					for (const [i, key] of keys.entries()) {
+						if (staleSet.has(key)) {
+							result[i] = undefined;
+						}
+					}
+
+					await this.deleteMany(staleKeys);
```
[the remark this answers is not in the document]
jaredwray: @AikidoSec ignore: false positive — `staleKeys` is a filtered subset of the caller-provided `keys: string[]` (typed cache keys), passed to Keyv's `deleteMany` which treats them as opaque, prefixed string keys. There is no query construction or operator interpretation in this path, so object/operator injection cannot occur. This mirrors the pre-existing public `deleteMany(keys)` API.

---

packages/cacheable/README.md
```diff
@@ -638,6 +639,63 @@ cache.maxTtl = '10m'; // 10 minutes max
...
+const writer = new Cacheable({ secondary: new KeyvRedis('redis://localhost:6379') });
+// instances that only read tagged entries should opt in with `tags: true`
+const reader = new Cacheable({ secondary: new KeyvRedis('redis://localhost:6379'), tags: true });
+
+await writer.set('page:/products', html, { tags: ['entity:42'] });
+await writer.tags.invalidateTag('entity:42');
+await reader.get('page:/products'); // undefined - stale copy is also purged from reader's primary
+```
+
+The tag service starts disabled so untagged workloads pay no extra cost. It enables automatically the first time a tag is written (setting a value with `tags` or calling `tags.invalidateTag` / `tags.invalidateTags`), or explicitly via the `tags: true` option or the `tags.enabled` property. When using tags across multiple instances that share a store, set `tags: true` on all of them so every instance honors invalidations and cleans up tag snapshots consistently.
```
jaredwray: this will not work in distributed instances. We should say you have to turn it on to use it. It should not auto turn on

jaredwray: Done in 90ec859 — auto-enable removed entirely. `CacheTags` never enables itself: while disabled, all operations (reads and writes) are no-ops, so values set with `tags` are stored without tracking and invalidations have no effect until you opt in. Docs now say you have to turn it on with `tags: true` (or `cache.tags.enabled = true`) on every instance that shares the store, and the examples all use `new Cacheable({ tags: true })`.

---

packages/cacheable/src/index.ts
```diff
@@ -638,6 +738,8 @@ export class Cacheable extends Hookified {
 				}
 			}
 
+			await this.setManyKeyTags(items);
```
jaredwray: check first if it is enabled

jaredwray: Done in 501cba1 — `setMany` now checks `this._tags.enabled` before calling `setManyKeyTags`, and the same explicit guard was added at the other call sites (`set`, `delete`, `deleteMany`) for consistency.

---

PR #1662, opened 2026-06-26
feat(memory): add statistics tracking via @cacheable/utils Stats

jaredwray:
## Summary

Adds opt-in runtime statistics to `@cacheable/memory` using the shared `Stats` implementation from `@cacheable/utils` — the same engine already used by `cacheable` and `@cacheable/node-cache`. This brings `CacheableMemory` in line with the rest of the suite for observability.

Statistics are **disabled by default**, so existing behavior is unchanged.

## What's included

- **New `stats` option** (`stats?: boolean`) on `CacheableMemoryOptions`, plus a `stats` getter that exposes the underlying `Stats` instance. Enable via the option or `cache.stats.enabled = true`.
- **Instrumented the leaf operations** (`get`, `getRaw`, `set`, `delete`, `clear`). Because the batch/derived methods delegate to these, `getMany`, `setMany`, `deleteMany`, `has`, `hasMany`, `take`, `takeMany`, and the `wrap`/`getOrSet` memoization helpers inherit tracking automatically.
- **Accurate size counters**: `count`/`ksize`/`vsize` are kept correct on overwrite (swap value size, don't double-count the key) and on delete, and reset on `clear`. LRU evictions are counted as deletes.
- **Re-exported** `Stats`, `StatsOptions`, and `StatsSnapshot` from the package entry.

## Counters tracked

`hits`, `misses`, `gets`, `sets`, `deletes`, `clears`, `count`, `ksize`, `vsize`, plus computed `hitRate`/`missRate` and a `toJSON()` snapshot.

```js
import { CacheableMemory } from '@cacheable/memory';

const cache = new CacheableMemory({ stats: true });
cache.set('key', 'value');
cache.get('key');     // hit
cache.get('missing'); // miss

cache.stats.hits;    // 1
cache.stats.misses;  // 1
cache.stats.hitRate; // 0.5
```

## Semantics note

Each individual key read counts as one `get` (so `getMany(['a','b'])` records two gets and two hits/misses), and `has()`/`take()` flow through `get`/`delete` and therefore update stats. This is documented in the README.

## Testing

- Added a `CacheableMemory Statistics` test suite covering enable/disable, every counter, overwrite, expiry-as-miss, LRU eviction, `take`/`has`/batch methods, `wrap`/`getOrSet`, `reset()`, and `toJSON()`.
- `pnpm --filter @cacheable/memory test` → **154 tests passing, 100% coverage** (statements/branches/functions/lines).
- `pnpm --filter @cacheable/memory build` succeeds.
- The `cacheable` consumer suite passes except for pre-existing Redis integration tests that require a Redis service (unavailable in this environment); those are unrelated to this change.

https://claude.ai/code/session_012LnVP8LHiL2fPi3BycCYWY

---
_Generated by [Claude Code](https://claude.ai/code/session_012LnVP8LHiL2fPi3BycCYWY)_

### DIFF
--- a/packages/memory/README.md
+++ b/packages/memory/README.md
@@ -294,6 +294,8 @@ You can get a plain-object snapshot via `cache.stats.toJSON()` and reset all cou
 
 The `count`, `ksize`, and `vsize` values are kept in sync as entries are added, removed, overwritten, and lazily expired, so they reflect the current contents of the cache. (Expired entries are not counted as `deletes`, since their removal is not user-initiated.) Methods that perform a read internally — such as `has()`, `take()`, and the `wrap()` / `getOrSet()` memoization helpers — flow through `get`/`set`, so they update the statistics as well.
 
+For accurate size counters, enable statistics before populating the cache: `count`/`ksize`/`vsize` only account for entries written while statistics were enabled, and are clamped at `0` so they never go negative if you enable stats after the cache already has data. Changing `storeHashSize` recreates the underlying stores and clears all entries, so the size counters are reset to `0` accordingly.
+
 ## CacheableMemory Options
 
 * `ttl`: The time to live for the cache in milliseconds. Default is `undefined` which is means indefinitely.
--- a/packages/memory/README.md
+++ b/packages/memory/README.md
@@ -29,6 +29,7 @@ Here are some of the main features of `CacheableMemory`:
 * [CacheableMemory Store Hashing](#cacheablememory-store-hashing)
 * [CacheableMemory LRU Feature](#cacheablememory-lru-feature)
 * [CacheableMemory Performance](#cacheablememory-performance)
+* [CacheableMemory Statistics](#cacheablememory-statistics)
 * [CacheableMemory Options](#cacheablememory-options)
 * [CacheableMemory - API](#cacheablememory---api)
 * [Keyv Storage Adapter - KeyvCacheableMemory](#keyv-storage-adapter---keyvcacheablememory)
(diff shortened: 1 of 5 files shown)


jaredwray:
Thanks for the review — good catch. Fixed in c0e1cdd.

Expired entries were being removed directly from the store on the lazy-deletion paths, so the size counters (`count`, `ksize`, `vsize`) never got decremented and drifted upward. I addressed it at **all** expiration-removal sites, not just `get()`/`getRaw()`:

- Added a private `recordExpiration(item)` helper that decrements `count`/`ksize`/`vsize` when stats are enabled.
- Called it in `get()`, `getRaw()`, the `keys` and `items` getters, and `checkExpiration()`.
- Expirations decrement the size stats but are **not** counted as `deletes`, since they aren't user-initiated.
- Also simplified the duplicate `item.expires && item.expires` condition in `getRaw()`.

Updated the `get`/`getRaw` expiry tests to assert the size stats return to `0`, and added tests for the `keys`/`items`/`checkExpiration` paths. Full suite: 157 passing, 100% coverage.

---

jaredwray:
Thanks @chatgpt-codex-connector — addressed all three findings.

1. **Decrement stats when expiring entries** (line 385): already handled in c0e1cdd (your review was against the initial commit `ae3ce30`). A `recordExpiration` helper now decrements `count`/`ksize`/`vsize` on every expiration-removal path — `get`, `getRaw`, the `keys`/`items` getters, and `checkExpiration` — without counting them as `deletes`.

2. **Counters go negative when stats are enabled late** (line 634): fixed in 14df090. `Stats.decreaseVSize`/`decreaseKSize`/`decreaseCount` now clamp at `0`, so deleting/overwriting an entry that was inserted before stats were enabled can no longer drive the counters negative. (Counters still only account for entries written while enabled, which is now documented — enable stats before populating for exact size totals.)

3. **`storeHashSize` change leaves stats stale** (line 540): fixed in 14df090. The `storeHashSize` setter recreates the stores (clearing all data), and now calls `resetStoreValues()` so `count`/`ksize`/`vsize` reset to `0` to match the emptied cache.

Added tests for the clamp behavior (`@cacheable/utils`) and the late-enable + `storeHashSize` paths (`@cacheable/memory`); verified `utils` (247), `memory` (159, 100% coverage), and `node-cache` (156) suites all pass.

---

PR #1664, opened 2026-06-26
feat: registry-aware release workflow with provenance + OIDC

jaredwray:
## Summary

Adds a robust, **fully tokenless** release pipeline that publishes **only** the workspace packages whose manually-set version is ahead of the npm registry. Versions are never bumped by the tooling — a human sets each `version` (e.g. in a release PR), and this decides what actually needs publishing and ships it in dependency order with provenance over OIDC.

Closes the gap visible in the repo today: the last release bump (8 packages, May 27) was merged but **never published** — the registry is still behind on all 8. This workflow publishes exactly those.

## What's included

- **`scripts/release.mjs`** — the decision + publish engine:
  - Enumerates publishable packages via `pnpm -r ls` (skips `private` packages and the internal `@cacheable/benchmark` harness).
  - Compares each local `version` against the npm registry: `already published → skip`, `new version / 404 → publish`. The **full plan is computed before any publish** — if a registry lookup can't be resolved (after retries + a 10s timeout), the run aborts before touching npm.
  - **Refuses to roll `latest` backward**: a version below the registry's `latest` dist-tag fails the run (a spec-correct semver comparator, no new dependency).
  - Publishes in **dependency (topological) order** — e.g. `@cacheable/utils` before `cacheable` — and **aborts on the first failure** so a failed dependency never leaves dependents pointing at a missing version.
  - **pnpm only**, with **provenance**: `pnpm --filter <name> publish --provenance --access public --no-git-checks`.
  - `--dry-run` prints the plan and validates packaging via `pnpm publish --dry-run`; `--json` emits the plan; writes a GitHub **step summary** table and `published-*` outputs.

- **`.github/workflows/release.yml`** — runs when a **GitHub release is published** (`release: [released]`, the same event as `deploy-website.yml`, so packages publish in lockstep with the website deploy) and on **manual dispatch** with a **Dry run** input. Grants `id-token: write` for **OIDC trusted publishing + provenance**, runs on a GitHub-hosted runner, `persist-credentials: false`. A manual dispatch from any ref other than `main` is **forced to a dry run**.

- **`CONTRIBUTING.md`** — a new **Releasing** section documenting the flow and the one-time npmjs.com trusted-publisher setup.

## Dry-run output (validated locally against the live registry)

```
package                local    registry  action
@cacheable/utils       2.4.2    2.4.1     PUBLISH
cacheable-request      13.0.19  13.0.19   SKIP
@cacheable/memory      2.1.0    2.0.9     PUBLISH
@cacheable/node-cache  3.1.0    3.0.1     PUBLISH
cache-manager          7.2.9    7.2.8     PUBLISH
cacheable              2.4.0    2.3.5     PUBLISH
@cacheable/net         2.0.9    2.0.8     PUBLISH
flat-cache             6.1.23   6.1.22    PUBLISH
file-entry-cache       11.1.4   11.1.3    PUBLISH
```

(Order above is the actual publish order — dependencies first.)

## ⚠️ One-time setup required before the first run

This workflow is **tokenless** — there is no `NPM_TOKEN` secret. For each published package, add a **Trusted Publisher** on npmjs.com (*Settings → Publishing access*): provider **GitHub Actions**, repo `jaredwray/cacheable`, workflow `release.yml`. Until a package has one configured, its publish step can't authenticate.

OIDC can't perform the *first-ever* publish of a brand-new package — bootstrap a new package's initial version once manually, then add its trusted publisher (documented in the **Releasing** section of `CONTRIBUTING.md`).

## Notes / decisions

- **`@cacheable/benchmark`** is `private: false` but has never been published and isn't in the documented public package list, so it's excluded via a documented `IGNORED_PACKAGES` constant. Remove it from that set to start publishing it.
- The release job builds + runs `scripts/test-build.mjs` (the publish-correctness gate) but does **not** run the full Redis test suite — `tests.yml` already gates `main`.
- A `release` event checks out the **tag**, so it publishes the versions in `package.json` at that tag — merge the version bump and point the release tag at it (normal flow: merge the bump PR → publish the release from `main`).
- This does **not** create git tags or GitHub Releases. If you'd like the workflow to also create per-package git tags, that's a quick follow-up.


.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+  # Manual runs — supports a dry run that prints the plan and validates
+  # packaging without publishing anything.
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: "Dry run (compute the plan + validate packaging, publish nothing)"
+        type: boolean
+        default: false
+  # A version bump always lands in a package.json, so only react to those.
+  push:
```
jaredwray: this should only be on publishing a release or workflow dispatch not push to main

.github/workflows/release.yml
```diff
@@ -0,0 +1,83 @@
...
+    branches: [main]
+    paths:
+      - "packages/*/package.json"
```
jaredwray: remove this

RELEASE.md
```diff
@@ -0,0 +1,61 @@
...
+- Workflow filename: `release.yml`
+
+The workflow is **fully tokenless** — there is no `NPM_TOKEN` secret. Every
+published package must have a trusted publisher configured before the workflow
+can publish it; until then that package's publish step fails to authenticate.
+
+One caveat: OIDC cannot perform the *first ever* publish of a brand-new package.
+When adding a new package, publish its initial version once manually (e.g.
+`pnpm --filter <name> publish --access public` while logged in to npm), then add
+its trusted publisher so every subsequent release flows through this workflow.
```
jaredwray: add this to the CONTRIBUTING.md instead of a new file

PR #1666, opened 2026-06-26
docs(net): expand @cacheable/net README with full API, Features, and Table of Contents

jaredwray:
## Summary

Makes the `@cacheable/net` README much more robust, mirroring the structure of the `cacheable` package README. It now documents the full public API surface with a complete, working Table of Contents.

### What changed
- **Expanded Features** list covering fetch semantics, method helpers, layered caching, RFC 7234, simple caching, coalescing, serialization, WHOIS/RDAP, Hookified, and typing.
- **Full Table of Contents** with anchors verified to resolve to real headings.
- **HTTP usage docs**: method helpers (`get`/`post`/`put`/`patch`/`delete`/`head`) with a signature/return/caching table, the `DataResponse` shape, `fetch` usage, caching control, RFC 7234 HTTP cache semantics, simple caching mode, custom serialization (`stringify`/`parse`), and error handling.
- **Custom/shared cache** section (construct from options or share a Layer 1/Layer 2 `Cacheable`), plus an **Events** section via the underlying cache.
- **Standalone functions** documented (`fetch`, `get`, `post`, `patch`, `del`, `head`) — including the note that there is no standalone `put` and why `del` is named `del`.
- **WHOIS & RDAP reference**: full `WhoisOptions` / `RdapOptions` / `QueryWhoisServerOptions` tables, `WhoisResult` / `WhoisHop` / `WhoisFields` / `RdapResult` type shapes, referral following, and the exported parsing helpers (`normalizeWhoisQuery`, `detectQueryType`, `parseWhois`, `queryWhoisServer`).
- **Reference sections**: `CacheableNet` options table, full API (properties + methods), and a complete exports/types list.

### Notes
- Documentation only — no source or behavior changes.
- API claims (signatures, options, defaults, type shapes, exported symbols), code examples, and every ToC anchor were cross-checked against `packages/net/src`.

https://claude.ai/code/session_01ERPCDjhDCjXjwrAkBD9z3Y

---
_Generated by [Claude Code](https://claude.ai/code/session_01ERPCDjhDCjXjwrAkBD9z3Y)_

### DIFF
--- a/packages/net/README.md
+++ b/packages/net/README.md
@@ -15,7 +15,7 @@
 * Optional response caching via [`cacheable`](https://npmjs.com/package/cacheable) — layered (Layer 1 / Layer 2) caching, LRU, TTL expiration, distributed sync, and more
 * [RFC 7234](http://httpwg.org/specs/rfc7234.html) compliant HTTP caching with `http-cache-semantics` (honors `Cache-Control`, `ETag`, `Last-Modified`, `Expires`, conditional revalidation, and `304 Not Modified`)
 * Simple TTL-based caching mode for when you don't want HTTP semantics (`httpCachePolicy: false`)
-* Smart, method-aware automatic cache key generation with request coalescing (concurrent identical misses share one request)
+* Smart, method-aware automatic cache key generation, with request coalescing in simple caching mode and for WHOIS/RDAP lookups (concurrent identical misses share one upstream request)
 * Custom serialization / deserialization with `stringify` and `parse` — at the instance level or per request
 * Request-level cache control with the `caching` option
 * `whois` and `rdap` lookups for domains, IPv4/IPv6 addresses, and ASNs with **both raw and JSON output**, dynamic IANA server discovery, referral following, and built-in caching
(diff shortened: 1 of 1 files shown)


jaredwray:
Thanks for the reviews. All three Codex points are addressed in 82a8ba5 (kept docs-only, since this PR is scoped to the README):

- **Write-method caching** — corrected to state that only **GET** (by default) and **PUT** (`caching: true`) are cached. `fetch.ts:128-136` skips the cache for POST/PATCH/DELETE/HEAD, so those are now documented as never cached, with a note that the `caching` option has no effect on them.
- **PUT cache identity** — the note now says the cache key is method + URL only (`fetch.ts:142`); the request body is not part of it.
- **Request coalescing** — scoped to simple caching mode (`httpCachePolicy: false`) and WHOIS/RDAP lookups; the default RFC 7234 path does not coalesce.
- Also clarified that the standalone `get`/`post`/`patch`/`head` helpers require an options argument and `del` requires an options object (they throw otherwise).

Note: the POST/PATCH/DELETE caching no-op (the `caching: true` option being silently ignored for those methods) looks like a latent bug in `fetch.ts` rather than intended behavior — flagging it here in case it's worth a separate code fix + tests that assert an actual cache hit.

---

PR #1670, opened 2026-07-05
mono - chore: upgrade build tooling

jaredwray:
- [x] Followed the [Contributing](../blob/main/CONTRIBUTING.md) guidelines and [Code of Conduct](../blob/main/CODE_OF_CONDUCT.md)
- [x] Tests for the changes have been added (for bug fixes/features) with 100% code coverage. — N/A: dependency-only upgrade with no source changes; existing suites still pass at 100% coverage where runnable in this environment.

**What kind of change does this PR introduce?**

Maintenance (`chore`) — upgrades the non-major **TypeScript / build tooling** to their latest `minimumReleaseAge`-gated versions. No source or public-API changes. Second PR of the dependency-management dev phase.

### Versions
- `tsdown` `^0.22.0` → `^0.22.3` (bundler; 8 packages: `cacheable`, `cache-manager`, `file-entry-cache`, `flat-cache`, `@cacheable/memory`, `@cacheable/net`, `@cacheable/node-cache`, `@cacheable/utils`)
- `tsx` `4.21.0` → `4.23.0` (`@cacheable/benchmark`, `@cacheable/website` — existing pin styles preserved: caret for benchmark, exact for website)
- `@arethetypeswrong/cli` `^0.18.3` → `^0.18.4` (root)

`typescript` (`5.9.3` → `6.0.3`) is a **major** and is deferred to its own PR; `@types/node` (`24.12.2` → `26.1.0`) is held pending a Node-major-version decision (`.nvmrc` says `22`, but CI tests `22/24/26` and the current pin is `^24`).

### Verification
- [x] `pnpm build` passes
- [x] `node scripts/test-build.mjs` (build-output validation: runtime + publint + attw) passes for all packages
- [x] `pnpm test` — every package passes at 100% coverage **except the same 3 tests** that fail on `main` and are specific to this sandbox (not caused by this change):
  - `cacheable-request › return with url and port` — real request to external `mockhttp.org:8080`, unreachable here (the `:80` variant passes).
  - `file-entry-cache › should return a file descriptor with checksum and error` and `normalizeEntries() › should return all entries` — both `chmod 0o000` a file to force a read error, but this sandbox runs as **root**, which bypasses permission bits.

  All three pass on GitHub CI (non-root runner, full network).

---
_Generated by [Claude Code](https://claude.ai/code/session_01P1TBuMiXotBjpNHYcbwvqf)_


jaredwray:
Re: Socket "Obfuscated code" alert on `@emnapi/runtime@1.11.1` — reviewed, no action needed on the code.

- `@emnapi/runtime` is a legitimate, widely-used Node-API/WASM runtime (part of the emnapi / napi-rs ecosystem). The "90% likely obfuscated" flag is a known heuristic false positive on its generated/minified WASM-interop glue, not malicious obfuscation.
- It was **already in the tree on `main`** (`@emnapi/runtime@1.9.2` and `@1.10.0`), pulled transitively via `wrangler → miniflare → sharp → @img/sharp-wasm32` and `@napi-rs/wasm-runtime`.
- This PR's only effect on it: the `tsdown` 0.22.0 → 0.22.3 patch bump transitively moves `@napi-rs/wasm-runtime` 1.1.4 → 1.1.6, which resolves `@emnapi/runtime@1.11.1` (up from 1.10.0) — same package, newer patch.
- It's a **dev/build-time transitive dependency only** (via the `tsdown` bundler and `wrangler`, both `devDependencies`); it is not a runtime dependency of any published `@cacheable/*` package.

The "Socket Security: Pull Request Alerts" check passed (this is a Warn-level advisory). I've deliberately not self-issued a `@SocketSecurity ignore`, since suppressing the alert is a maintainer triage decision — happy to add it or pin differently if you'd prefer.

---

